PR #136, opened 2017-09-05
Flekschas/brushing

flekschas:
Massive monster pull request 😬

The pull request consists of the following main changes:
- Add API-based brush-based range selection
- Extended API with listener for range selection and getter for all events
- Introduce ESLint and apply auto-formatting to all files
- Introduce CSS Modules to avoid future CSS conflicts when integrating HG in another app
- Lots of code clean-up (removed dead code and added missing dependencies)


pkerpedjiev:
close track menu looks a little strange:

image

pkerpedjiev:
If a view is made too narrow, the toolbar icons stick out of the side:

image

In such situations, the search area should disappear and leave only the control icons.

flekschas:
Agreed. I'll think of a way hide it gracefully

flekschas:
@pkerpedjiev 
I've fixed the issue with the closing menu and hide the search bar under 300px view header width

Screenshot at 150px window width:
screen shot 2017-09-05 at 7 08 37 pm

PR #270, opened 2018-06-15
I think this is everywhere that same-origin is required.

mccalluc:
Not sure about the behavior of d3-requests.

I was a little confused by worker.js, but this looks like the appropriate translation if we're adding more to the init: (Request and fetch have the same signature.)
```javascript
headers = {'content-type': 'application/json'};
r = new Request('http://example.com', {headers});
console.log('Headers:', Array.from(r.headers.keys()));
console.log('Credentials:', r.credentials);
// Headers: Array [ "content-type" ]
// Credentials: omit 

headers = {'content-type': 'application/json'};
r = new Request('http://example.com', {'headers': headers, 'credentials': 'same-origin'});
console.log('Headers:', Array.from(r.headers.keys()));
console.log('Credentials:', r.credentials);
// Headers: Array [ "content-type" ]
// Credentials: same-origin
```

### DIFF
--- a/app/scripts/HiGlassComponent.js
+++ b/app/scripts/HiGlassComponent.js
@@ -2319,7 +2319,8 @@ class HiGlassComponent extends React.Component {
           'Content-Type': 'application/json',
           'X-Requested-With': 'XMLHttpRequest',
         },
-        body: `{"viewconf":${this.getViewsAsString()}}`
+        body: `{"viewconf":${this.getViewsAsString()}}`,
+        credentials: 'same-origin',
       }
     )
       .then( response => {
--- a/app/scripts/services/chrom-info.js
+++ b/app/scripts/services/chrom-info.js
@@ -14,7 +14,7 @@ const parseChromInfo = (text) => {
   return parseChromsizesRows(tsv);
 };
 
-const getFromRemote = url => fetch(url)
+const getFromRemote = url => fetch(url, {credentials: 'same-origin'})
   .then(response => response.text())
   .then(text => parseChromInfo(text))
   .catch((error) => {
--- a/app/scripts/services/tile-proxy.js
+++ b/app/scripts/services/tile-proxy.js
@@ -552,6 +552,7 @@ function json(url, callback) {
 
   const r = request(url)
     .header('Content-Type', 'application/json')
+  // TODO: Check if this preserves same-origin cookies
 
   if (authHeader)
     r.header('Authorization', `${authHeader}`)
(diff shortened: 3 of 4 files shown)


mccalluc:
ping @pkerpedjiev: Could this be merged? Background: I'm not sure why things seemed to work in the past, but I'm pretty confident that going forward this is what is required for higlass to work inside refinery.

 (cc @scottx611x)

pkerpedjiev:
Isn't `same-origin` the default value for `credentials`? https://developer.mozilla.org/en-US/docs/Web/API/Request/credentials

Have you tried building a local version of HiGlass and seeing if this PR solves the issue you have?

mccalluc:
(Copying thread from slack for reference:)

huh: I’m trying to re-run the experiment in the PR, but now it looks like this:
```javascript
headers = {'content-type': 'application/json'};
// Object { "content-type": "application/json" }
r = new Request('http://example.com', {headers});
// Request { method: "GET", url: "http://example.com/", headers: Headers, destination: "", referrer: "about:client", referrerPolicy: "", mode: "cors", credentials: "same-origin", cache: "default", redirect: "follow" }
r.credentials
// "same-origin"
```
I’m wondering if I misread some part of the docs once, and then never checked that belief? … but I’m also pretty sure cookies were being dropped. Let me look into this again and get back to you.

mccalluc:
browser might make a difference:

Chrome 67.0.3396.99:
```
headers = {'content-type': 'application/json'};
r = new Request('http://example.com', {headers});
console.log('Headers:', Array.from(r.headers.keys()));
console.log('Credentials:', r.credentials);
// Headers: ["content-type"]
// Credentials: omit
```

FF 61.0:
```
headers = {'content-type': 'application/json'};
r = new Request('http://example.com', {headers});
console.log('Headers:', Array.from(r.headers.keys()));
console.log('Credentials:', r.credentials);
// Headers: Array [ "content-type" ]
// Credentials: same-origin
```

I think [this](https://www.chromestatus.com/feature/4539473312350208) is the issue in Chrome to fix this behavior, and [this](https://github.com/whatwg/fetch/commit/811575d6889026c68d55a327f79198dccfea403c) is the change in the spec they refer to.

Let me:
- See if I can find a corresponding recent change in FF.
- Run a local demo, which I believe would show the right behavior in FF, but the wrong behavior in Chrome.

mccalluc:
... or, actually, that would be a bit of a pain... Locally edit this repo, and then build gehlenborglab/higlass pointing to local edits, and then build the refinery wrapper, modified to point to the local base image... What do you think? Is this a scenario we should find a way to simplify?

PR #278, opened 2018-06-28
Flekschas/enhanced bed track

flekschas:
This PR adds color encoding to the bed track

Here's an example config:

```
{
  name: 'loss',
  server: '//localhost:8001/api/v1',
  tilesetUid: 'loss',
  uid: 'loss',
  type: 'bedlike',
  options: {
    fillColor: 'purple',
    axisPositionHorizontal: 'right',
    labelColor: 'black',
    labelPosition: 'topLeft',
    trackBorderWidth: 0,
    trackBorderColor: 'black',
    labelTextOpacity: 0.4,
    valueColumn: 4,
    colorEncoding: true,
    colorRange: [
      '#000000', '#652537', '#bf5458', '#fba273', '#ffffe0'
    ],
    colorEncodingRange: [0, 0.5119949],
    fillOpacity: 0.9,
    name: 'loss',
    showMousePosition: true,
    mousePositionColor: '#000000',
  },
  height: 24,
  position: 'top'
}
```

Right now one can either encode values in `y` (with `colorEncoding: false`, which is the default value) or as `color`. Both at the same time is not implemented yet.


pkerpedjiev:
A little docs?

flekschas:
@pkerpedjiev What do you think: https://github.com/hms-dbmi/higlass/pull/278/commits/7ebd07b3b3a837859481db8ee93ac3765a5bb8c6

PR #285, opened 2018-07-17
Flekschas/1d annotation track

flekschas:
A wonderful simple 1D (horizontal + vertical) generic annotation track. Can be used for static highlighting of a certain region.

Example [app/test-1d-annotations.html](app/test-1d-annotations.html)


pkerpedjiev:
Is there documentation for this? If I had to use it, what would I have to do?

PR #289, opened 2018-07-26
Flekschas/reset viewport api

flekschas:
To reset the viewport to the initial x and y domains do the following:

```javascript
const hgApi = hglib.viewer(document.getElementById('demo'), viewConfig, { bounded: true });
hgApi.resetViewport();
```

While HiGlass is working hard on accomplishing your demands, grab a nice cup of ☕️, enjoy the 🏝, and call it a week! Once you're back at work (or after 2ms) the viewport will be reset to it's beautiful original state.

### DIFF
--- a/app/scripts/HiGlassComponent.js
+++ b/app/scripts/HiGlassComponent.js
@@ -1064,6 +1064,25 @@ class HiGlassComponent extends React.Component {
     }
   }
 
+  /**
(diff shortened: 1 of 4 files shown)


pkerpedjiev:
Could you create a PR to `higlass-docs` with a description of this API endpoint?

Issue #287, opened 2018-07-20
same-origin cookies not preserved in tile-proxy.js#json() in Chrome


scottx611x:
_Commit_: https://github.com/hms-dbmi/higlass/commit/9b6592abab5b2662a4599af2d838592d97470888

#### Steps to reproduce
1.  Load a track in a recent version of Chrome

#### Observed behavior
* Note how the `d3-request` [above this TODO](https://github.com/hms-dbmi/higlass/blob/30bc4779f4a7892293a092965a7b7025e2f21a22/app/scripts/services/tile-proxy.js#L556) does not preserve `same-origin` cookies.

#### Expected behavior
- For cookies to be preserved

Take note that recent versions of d3 utilize the `Fetch API` which ["doesn't send any cookies by default"](https://stackoverflow.com/a/50726741)

mccalluc:
I believe we fixed this: The TODO is gone, and it looks like the code is jumping through the right hoops to keep credentials: 
```
return fetch(url, { credentials: 'same-origin', headers })
```

We can reopen or file a new issue if there are still problems.

PR #308, opened 2018-08-13
switch to Fetch API

mccalluc:
Towards #287.
- I probably should have branched from develop, right?
- How do I test things on this project?
- The two methods are now almost identical, except for `json` and `text` on one line, and the mime type: Introduce a wrapper method?

### DIFF
--- a/app/scripts/services/tile-proxy.js
+++ b/app/scripts/services/tile-proxy.js
@@ -532,36 +532,54 @@ export const tileDataToPixData = (
 
 function text(url, callback) {
   /**
-   * Send a JSON request mark it so that we can tell how many are in flight
+   * Send a text request and mark it so that we can tell how many are in flight
    */
   requestsInFlight += 1;
   pubSub.publish('requestSent', url);
 
-  d3Text(url, (error, done) => {
-    callback(error, done);
-    pubSub.publish('requestReceived', url);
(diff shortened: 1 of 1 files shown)


app/scripts/services/tile-proxy.js
```diff
@@ -532,36 +527,54 @@ export const tileDataToPixData = (
...
+      if (response.ok) {
+        response.text().then(t => {
+          callback(null, t);
+          requestsInFlight -= 1;
+        })
+      } else {
+        // Do something?
+      }
+    })
+    .catch(error => {
```
flekschas: Flight counter is not updated. I'd suggest using `finally` to update it.

flekschas: I am done a super similar update here https://github.com/hms-dbmi/higlass/pull/307
Maybe we can merge the two efforts. :)

flekschas:
This is very likely going to clash with https://github.com/hms-dbmi/higlass/pull/307 where I replace d3text with `fetch`. Maybe we can merge the two efforts :)

app/scripts/services/tile-proxy.js
mccalluc: How about if I make a branch off of your branch, and just duplicate what you've done for `text` for `json`? Thanks!

PR #310, opened 2018-08-14
Mccalluc/add json to update chrominfo zoom to

mccalluc:
I believe this is want we want for fetching so that 3rd-party cookies will be preserved, but I'm getting a number of errors when I run tests locally, even if I try to run from develop. I also don't have a good test of the failure callbacks. Travis seems to run only on PRs?


mccalluc:
Note that although travis is happy, tests aren't actually being run:
```
0.51s$ npm test
> higlass@1.1.5 test /home/travis/build/hms-dbmi/higlass
> echo 'Error: no test specified'
Error: no test specified
The command "npm test" exited with 0.
```

app/scripts/services/tile-proxy.js
```diff
@@ -533,43 +528,45 @@ export const tileDataToPixData = (
...
-      callback(undefined, text);
-    })
-    .catch((error) => {
-      callback(error, undefined);
-    })
-    .finally(() => {
-      pubSub.publish('requestReceived', url);
-      requestsInFlight -= 1;
-    });
+  fetchEither(url, callback, 'text');
```
flekschas: `return` is missing. Callbacks are old-school but we need them for compatibility but we should also return the promise such that we can progressively update the code base.

app/scripts/services/tile-proxy.js
```diff
@@ -533,43 +528,45 @@ export const tileDataToPixData = (
...
-  // TODO: Check if this preserves same-origin cookies
-
-  if (authHeader)
-    r.header('Authorization', `${authHeader}`)
+  if (textOrJson == 'text') {
+    var mime = 'text/plain';
+  } else if (textOrJson == 'json') {
+    var mime = 'application/json';
+  }
+  var headers = {'Content-Type': mime};
```
flekschas: `var` is ES5 and outdated. `const` is preferred here.

app/scripts/services/tile-proxy.js
```diff
@@ -533,43 +528,45 @@ export const tileDataToPixData = (
...
   requestsInFlight += 1;
   pubSub.publish('requestSent', url);
 
-  const r = request(url)
-    .header('Content-Type', 'application/json')
-  // TODO: Check if this preserves same-origin cookies
-
-  if (authHeader)
-    r.header('Authorization', `${authHeader}`)
+  if (textOrJson == 'text') {
```
flekschas: Not crucial but in the JS world always use `===` instead of `==`. Often they are equivalent but sometimes funky things can happen, so it's recommended to just stick to `===`.

mccalluc:
Tried to rebase and I think I just confused things: @pkerpedjiev : Would you prefer that I make a new clean PR this, even if the tests pass?

pkerpedjiev:
Yes please.

mccalluc:
replaced by #310

PR #311, opened 2018-08-14
Mccalluc/just add test runner tweaks

mccalluc:
Not really a Pull "Request" yet, but travis doesn't run except for PRs: Don't feel obliged to look at it, yet. The best case would be to see the same errors on travis that I see locally.


mccalluc:
yay! Travis is actually giving me the same errors I see locally: I'll look at this a little bit more and let you know what I learn.

mccalluc:
@pkerpedjiev : Can you confirm that there's nothing useful in this branch, and I can just close and delete it?

pkerpedjiev:
Close and delete it!

PR #314, opened 2018-08-16
Flekschas/enhance gene annotations

flekschas:
Allow changing the font size of gene labels, the height of gene rects, spacing between the two strands, and drawing gene labels inside the gene rects. The whole point is to save space. This reduced the min height from 55px to 24px (default is still 55px).

Comparison between customization (top, black) and the default look (middle, red-blue):

screen shot 2018-08-16 at 1 50 58 pm

With these abilities it's also easy to _hide_ the gene labels entirely (bottom, purple-orange)


pkerpedjiev:
Is it OK if we change `geneAnno...` to `geneAnnotation...`?

flekschas:
No way! 😤

... of course 😇😄. Are you referring to `geneAnnoHeight`?

pkerpedjiev:
Thanks!

> Are you referring to geneAnnoHeight?
Yes.

PR #315, opened 2018-08-16
preserve cookies in fetch

mccalluc:
Following the instructions in README does not run tests for me, right now. Looking into it...


mccalluc:
(Do we know if the `text` method is ever used? Trying to track down usages...)

mccalluc:
5 failing test, all involving "TypeError: Cannot read property 'getTrackObject' of null".

app/scripts/services/tile-proxy.js
```diff
@@ -533,45 +528,48 @@ export const tileDataToPixData = (
...
-
-    r.send('GET', (error, data) => {
-      pubSub.publish('requestReceived', url);
-      const j = data && JSON.parse(data.response);
-      callback(error, j);
-      requestsInFlight -= 1;
-    });
+  
+  if (textOrJson === 'text') {
+    var mime = 'text/plain';
```
pkerpedjiev: No `var` please. Use `let` or `const`.

app/scripts/services/tile-proxy.js
```diff
@@ -533,45 +528,48 @@ export const tileDataToPixData = (
...
-      pubSub.publish('requestReceived', url);
-      requestsInFlight -= 1;
-    });
+  return fetchEither(url, callback, 'text');
 }
 
 function json(url, callback) {
   /**
-   * Send a JSON request mark it so that we can tell how many are in flight
+   * Send a JSON request and mark it so that we can tell how many are in flight
```
pkerpedjiev: Any chance we could get some @param tags?

mccalluc:
@pkerpedjiev : 
- I don't think the failures are related to changes here, but it'd still be nice if tests could pass.
- When this is done, we still need to pin a release, and update 2 docker images, I think. I guess let me know when there's a tag I can reference, and I'll make another PR for the docker image, unless you want to tackle that?

PR #330, opened 2018-10-23
make zoomTo example non-degenerate

mccalluc:
... but these docs don't actually match [docs-dev.higlass.io](http://docs-dev.higlass.io/javascript_api.html#zoomTo). There is also [api.js](https://github.com/higlass/higlass/blob/2f1b7b4b38fe3a706381b4b41bcc00aa9ad00a72/app/scripts/api.js#L397) ... and I guess I'll make a similar edit there. [higlass-docs](https://github.com/higlass/higlass-docs/blob/develop/_build/html/javascript_api.html#L391) doesn't seem like the right place to edit... I wouldn't think the built html should even be checked in?

### DIFF
--- a/docs/higlass_developer.rst
+++ b/docs/higlass_developer.rst
@@ -231,9 +231,9 @@ to each other.
 ``end1: Number``
     The right x coordinate of the region to zoom to.
 ``start2: Number``
-    The left x coordinate of the region to zoom to.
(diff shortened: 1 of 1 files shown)


pkerpedjiev:
Sorry, I think I mislead you earlier today. The api documentation is auto-generated from the javadoc strings in `api.js`. The other PR should be enough.

PR #334, opened 2018-10-25
Mccalluc/get svg working

mccalluc:
When I copy and paste one test, it fails with `Unable to find node on an unmounted component.` Maybe see if it works for you, or if it's something about my environment.

(I'm happy with everything except the testing, though criticism is welcome!)

### DIFF
--- a/test/PngExportTest.js
+++ b/test/PngExportTest.js
@@ -2,110 +2,66 @@
 import {
   configure,
   mount,
-  // render,
-  ReactWrapper,
 } from 'enzyme';
 
 import Adapter from 'enzyme-adapter-react-16';
 
 import { expect } from 'chai';
-import { select } from 'd3-selection';
-import React from 'react';
-import ReactDOM from 'react-dom';
-import slugid from 'slugid';
 
 import HiGlassComponent from '../app/scripts/HiGlassComponent';
-import HeatmapOptions from '../app/scripts/HeatmapOptions';
 
 // Utils
 import {
-  scalesCenterAndK,
-  totalTrackPixelHeight,
-  getTrackByUid,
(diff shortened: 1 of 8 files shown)


test/PngExportTest.js
```diff
@@ -0,0 +1,111 @@
...
+  getTrackByUid,
+  getTrackObjectFromHGC,
+  getTiledPlot,
+  waitForJsonComplete,
+  waitForTilesLoaded,
+  waitForTransitionsFinished,
+} from '../app/scripts/utils';
+
+// View configs
+import {
```
pkerpedjiev: No need to import all these viewconfigs if only one is being used, right?

test/PngExportTest.js
```diff
@@ -0,0 +1,111 @@
...
+import { select } from 'd3-selection';
+import React from 'react';
+import ReactDOM from 'react-dom';
+import slugid from 'slugid';
+
+import HiGlassComponent from '../app/scripts/HiGlassComponent';
+import HeatmapOptions from '../app/scripts/HeatmapOptions';
+
+// Utils
+import {
```
pkerpedjiev: Are all these imports being used?

test/PngExportTest.js
```diff
@@ -0,0 +1,111 @@
...
+      hgc = mount(<HiGlassComponent
+        options={{ bounded: false }}
+        viewConfig={testViewConfX1}
+      />, { attachTo: div });
+
+      hgc.update();
+      waitForTilesLoaded(hgc.instance(), done);
+    });
+
+    it ("Exports to SVG", (done) => {
```
pkerpedjiev: There's no test for PNG export in here? I suppose you were planning on updating it later.

Issue #337, opened 2018-10-26
Clean up copy-and-paste between HorizontalRule, VerticalRule, and CrossRule


mccalluc:
(The rules are not being exported in either the SVG or PNG, so that's why I started looking at these.)

Horizontal | Vertical | Cross | ?
--- | --- | --- | ---
`drawHorizontalRule` | `drawVerticalRule` | | Style code redundant, draw code reverses coordinates
`mouseMoveHandler` | `mouseMoveHandler` | | Has diverged in unintended way?
`drawHorizontaRule` | | `drawHorizontalRule`| copy and paste
`isMouseOverHorizontalLine` | | `isMouseOverHorizontalLine` | copy and paste

PR #340, opened 2018-10-29
Export Horizontal and Vertical rules

mccalluc:
CrossRule and Vertical work, but not Horizontal: it gets PixiTrack's exportSVG rather than our method.

There's also boiler plate in these methods that I don't understand... why do we return a two element list? Maybe I should tackle #337 first? Also needs testing.


pkerpedjiev:
I'm actually having a hard time remembering why we have that boiler plate. From what I can remember and infer from the code it has to do with the track class hierarchy. Things like labels are drawn by PixiTrack while the actual track-specific visualizations are drawn by the specific tracks (e.g. HeatmapTiledPixiTrack). 

Since the output is a tree, the root is `gBase`. Adornment like labels are drawn on child branch of `gBase`. The other branch is reserved for track-specific drawing and returned as `gTrack`.

pkerpedjiev:
Do you have the code you tried for the `HorizontalRule`? I'm curious why it doesn't behave like the vertical.

mccalluc:
Closing this: It's not something I need, or that I plan to work on.

PR #346, opened 2018-10-30
Mccalluc/get examples working

mccalluc:
These examples were not working, and now I think they are.


pkerpedjiev:
Looks good. Thanks!

PR #347, opened 2018-10-30
Mccalluc/get examples working in new directory

mccalluc:
I'm not sure what the documentation site expects, but making the paths absolute and adding a symlink dir lets the examples work with the local development server... Does this seem like the right direction, and will it work for documentation site?


mccalluc:
Both builds failed on `Simple HiGlassComponent Viewport projection tests Dispatches a mousewheel event on the viewport projection and makes sure it zooms FAILED`: Restarting travis builds.

pkerpedjiev:
That build error has been happening for a while, it's unrelated to your change. I've working on a different bug but will swing around to try to fix it.

I think that's fine for the examples for now but perhaps @flekschas could weigh in? Fritz, when the examples are in `docs/examples` how did you run them in development mode?

flekschas:
@mccalluc I think I am missing some context. Why exactly are you changing the paths? Everything worked for me last time I've checked but maybe something changed?

> how did you run them in development mode?

`npm start` and then `http://localhost:123(I forgot the port number)/example-name.html`

mccalluc:
Oh! I just didn't realize there was routing across the directories, but now I see `path.resolve(__dirname, 'docs', 'examples')` in webpack.config.js, and I think that's what does it?

I think it might help to navigate these if there was a link from the top level index.html to the examples directory?

But for now, I'll get rid of the absolute paths and directory symlink I added here, and then this PR should be ready. Thanks, @flekschas!

mccalluc:
@pkerpedjiev : Is this ok now?

PR #348, opened 2018-10-31
Mccalluc/cnv explorer working

mccalluc:
I'll make a separate branch to work on the overlay infrastructure, but I think this much works, and could be merged now. Not sure if all the data files are actually being used? After merge https://github.com/pkerpedjiev/cnv-explorer could be deleted, or maybe just add a note in the readme pointing to this.


mccalluc:
- You're right, `K562_regions.tsv.multires.db` is not needed.
- neither is `7eb0f792-....somatic.cna.annotated.txt`? It seems to work without that one. Removed.
- I had plain `hglib.js` at first, and then decided it would be more clear if the html and the data were in the same directory together... so now it's down a level, and it would either need `../` to make the relative path work, or I can make it absolute, and I think absolute is more clear?

mccalluc:
@pkerpedjiev : how about now?

PR #354, opened 2018-11-01
Refactor some of the tests

pkerpedjiev:
* Added the `removeHGComponent` helper method for unmounting the HiGlass component and removing the div which contains it after a test
* Added a test for zooming to regions with negative coordinates

### DIFF
--- a/app/scripts/HiGlassComponent.js
+++ b/app/scripts/HiGlassComponent.js
@@ -3057,7 +3057,7 @@ class HiGlassComponent extends React.Component {
     }
 
     if (
-      !(+start1Abs >= 0 && +end1Abs >= 0)
+      isNaN(start1Abs) || isNaN(end1Abs)
     ) {
       const coords = [start1Abs, end1Abs].join(', ');
       console.warn([
--- a/app/scripts/utils/test-helpers.js
+++ b/app/scripts/utils/test-helpers.js
@@ -195,3 +195,10 @@ export const mountHGComponent = (prevDiv, prevHgc, viewConf, done, options) => {
 
   return [div, hgc];
 };
+
+export const removeHGComponent = (div) => {
+  // console.log('removing', div);
+
+  ReactDOM.unmountComponentAtNode(div);
+  document.body.removeChild(div);
+}
--- a/test/APITests.js
+++ b/test/APITests.js
@@ -12,6 +12,7 @@ import { expect } from 'chai';
 import {
   waitForTransitionsFinished,
   waitForTilesLoaded,
+  removeHGComponent,
 } from '../app/scripts/utils';
 
 import {
@@ -45,7 +46,7 @@ describe('Simple HiGlassComponent', () => {
       // console.log('viewHeaders:', component.viewHeaders);
       expect(Object.keys(component.viewHeaders).length).to.be.above(0);
 
-      document.body.removeChild(div);
+      removeHGComponent(div);
     });
 
     it('zooms to negative domain', (done) => {
@@ -57,8 +58,7 @@ describe('Simple HiGlassComponent', () => {
 
       waitForTransitionsFinished(api.getComponent(), () => {
         expect(api.getComponent().yScales.a.domain()[0]).to.be.below(0);
-        document.body.removeChild(div);
-
+        removeHGComponent(div);
         done();
       });
     });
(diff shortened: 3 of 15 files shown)


app/scripts/HiGlassComponent.js
```diff
@@ -3057,7 +3057,7 @@ class HiGlassComponent extends React.Component {
     }
 
     if (
-      !(+start1Abs >= 0 && +end1Abs >= 0)
+      isNaN(start1Abs) || isNaN(end1Abs)
     ) {
       const coords = [start1Abs, end1Abs].join(', ');
       console.warn([
```
mccalluc: FWIW, I think the backtick syntax can span multiple lines.

app/scripts/utils/test-helpers.js
```diff
@@ -195,3 +195,10 @@ export const mountHGComponent = (prevDiv, prevHgc, viewConf, done, options) => {
 
   return [div, hgc];
 };
+
+export const removeHGComponent = (div) => {
+  // console.log('removing', div);
```
mccalluc: This commented out bit probably isn't useful? Unless you expect to continue debugging?

test/APITests.js
```diff
@@ -71,20 +70,15 @@ describe('Simple HiGlassComponent', () => {
...
-          // console.log('trackObj', trackObj);
-          // trackObj.getMouseOverHtml(50,50);
 
           const rd = trackObj.getVisibleRectangleData(285, 156, 11, 11);
           expect(rd.data.length).to.eql(1);
-          // trackObj.getVisibleRectangleData(24,8,11,11);
 
-          // console.log('visi', trackObj.visibleAndFetchedTiles());
-          document.body.removeChild(div);
+          removeHGComponent(div);
```
mccalluc: Here and below, maybe put it in an afterEach, rather than needing to remember to include it on every test. (And if every test wouldn't use the same afterEach, rearrange so they would?)

test/HiGlassComponentTest.js
```diff
@@ -3133,4 +3133,21 @@ describe('Simple HiGlassComponent', () => {
     });
 
   });
+
+  
+  describe('Cleanup', () => {
```
mccalluc: Maybe use an `after` instead, since this isn't really a test?

test/OSMTests.js
```diff
@@ -6,12 +6,11 @@ import {
 
 import Adapter from 'enzyme-adapter-react-16';
 
-import { expect } from 'chai';
+// import { expect } from 'chai';
```
mccalluc: Can we just delete?

test/OSMTests.js
```diff
@@ -46,19 +45,17 @@ describe('Simple HiGlassComponent', () => {
       osmConf.views[0].initialYDomain = [-42.4, -42.3];
       osmConf.views[0].layout.w = 12;
 
-      // console.log('track:', track);
       const [div, api] = createElementAndAPI(osmConf, {
         bounded: true
       });
 
-      const component = api.getComponent();
+      // const component = api.getComponent();
```
mccalluc: delete?

pkerpedjiev:
> For tests, I think it's generally good if they are insensitive to order, and use before/beforeEach/after/afterEach where you do need setup or cleanup.

Yeah, I agree. A lot of these tests were written before I figured out how to quickly instantiate a component. I think in the future we should aim for `beforeEach` and `afterEach` but for now we can leave all the existing tests be.

> Linting: what tool do you use? My editor wanted to take out extra spaces on save, but that introduced a lot of whitespace changes... Could we clean it up once, and then include linting as part of the test suite to try to keep it tidy?

I use `eslint` as part of SublimeLinter. In general I would agree, but cleaning up the existing code is going to be a monumental effort and there's some rules that are useful but shouldn't be enforced universally (e.g. `beforeAll` not defined in tests). Let's stick with clean as can be code going forward and fixing whatever is in context instead of spending the time to go and try to fix everything at once.

app/scripts/utils/test-helpers.js
pkerpedjiev: removed

app/scripts/HiGlassComponent.js
pkerpedjiev: Oh, good to know!

test/APITests.js
pkerpedjiev: Good idea. Changed.

test/OSMTests.js
pkerpedjiev: I added a test that uses this, so... uncommented rather than deleted :)

test/HiGlassComponentTest.js
pkerpedjiev: Leaving this one be because that test file is a huge can of worms that I don't want to disturb.

PR #355, opened 2018-11-01
Mccalluc/overlay example

mccalluc:
I need to write tests, and there's an offset in the masking that I don't understand, but the demo looks good.

(This still has each overlay based on a single extent, but I think you wanted that changed to a list of extents?)


mildewey:
#538 fixes extent only accepting 1 input and also allows for 4-tuples as an extent range. There will definitely be conflicts, though, as it also restructures some of the code I see changed in this PR.

mccalluc:
Closing this: It's not something I need, or that I plan to work on.

PR #356, opened 2018-11-01
Mccalluc/install eslint

mccalluc:
eslint was missing dependencies, but this works now for me. I'm guessing you guys had installed globally and so you were getting all the dependencies from there? To make sure I've got all the dependencies, I wanted to run it against something, but picking karma.conf.js is arbitrary.

(If eslint should have worked out of the box, then ignore this, and I can try something different to get it working locally.)


pkerpedjiev:
Could you merge this with the latest develop branch to make sure the tests pass?

mccalluc:
passed locally.

pkerpedjiev:
Thanks! Yeah, I have eslint installed globally so that it works with SublimeText. 

I like how you have it run on karma.conf.js. We can start adding other files as they get cleaned up.

PR #358, opened 2018-11-05
Mccalluc/simplify docs deploy

mccalluc:
I think I outlined this to Pete earlier; Here's the PR.

My sense is that with the current documentation system, there is really only one goal (separate master and develop doc sites) and everything else (git subtree, ghpages, multiple repos) arises from that. I think this is a simpler approach that still does what you need. If I've misunderstood your needs, let me know, and I can see if this can be adapted.

I haven't set up a CNAME, but the [site is up](http://higlass-docs-demo.s3-website-us-east-1.amazonaws.com/)

Features:
-  Successful `master` branch pushes to `/`; `develop` pushes to `develop`; and all others push to [`latest`](http://higlass-docs-demo.s3-website-us-east-1.amazonaws.com/latest/)
- Weirdness with git subtree and twinned docs repos no longer needed: The docs source can actually live in this repo. (I'm not really clear on what is needed to undo a subtree, and just have the files in-place.)
- Doc build artifacts are no longer checked in.

Caveats:
- Needs a CNAME.
- I've done this under the Refinery AWS account, because that's what I have access to.
- The "master" deployment up now is not actually `master`: I just pushed one build to the top level before I got the branch check implemented.
- Nothing under `/develop` right now... but there should be when we merge to develop.
- Having the separate builds as subdirectories is optional: We could give each its own bucket... just seemed like more trouble, but not a strong preference.

Thoughts?


pkerpedjiev:
In general, I think this is great! 

One concern is that the docs currently also contain documentation for `higlass-jupyter` as well as for `hgflask`. It might be strange to have those be built according to the higlass client release schedule. 

A possible solution is to have those repositories maintain their own docs folders and push to sub-paths of the AWS s3 bucket.

When you say pushes to `develop`, do you mean that it copies the documentation to here: http://higlass-docs-demo.s3-website-us-east-1.amazonaws.com/develop/?

Just out of curiosity, why do you use `pushd` and `popd` rather than `cd` and `cd -`?

Issue #16, opened 2017-01-31
Use a logging library


flekschas:
_Commit_: cafdba59fd0842cc279b8ad3708c8a955bfc872a

#### Steps to reproduce
1. Open the console

#### Observed behavior
Lot's of logging is going on.

#### Expected behavior
HGLib shouldn't log anything unless it is launched in some sort of debug mode. Logging is usually not a problem but in case somebody integrates HiGlass in a different app (like I do) than having tons of extra logs is a little but inconvenient.

flekschas:
Since it's easy to forget about loggers and they are helpful during development I've personally got into the habit of allowing them in local dev mode but cause an ESLint error when compiling in production. This lets you also easily catch issues with Travis before publishing a new version.

PR #359, opened 2018-11-06
Mccalluc/more linting

mccalluc:
Add app/scripts/services to lint coverage. This is obviously not a priority, but... 
- It's a prompt for me to skim a lot of code.
- It helps me internalize coding practices on this project.
- and it adds a few files to the set of those that will be linted on each push.

If the fix is straight-forward, I've gone ahead a made it... but if it seemed weird, I've added an ignore, rather that superficially cleaning up the syntax, and potentially hiding a deeper problem.

Thoughts?

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -36,7 +36,7 @@
     "guard-for-in": 0,
     "jsx-quotes": 1,
     "no-bitwise": 0,
-    "no-console": 0,
+    "no-console": ["error", { "allow": ["warn", "error"] }],
     "no-continue": 0,
     "no-mixed-operators": ["error", { "allowSamePrecedence": true }],
     "no-multi-spaces": ["error", { "ignoreEOLComments": true }],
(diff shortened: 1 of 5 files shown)


pkerpedjiev:
It looks like the only rule you added was:

```
    "no-console": 0,
```

right?

mccalluc:
Yep, that's the one change to the config. My sense is that in the long run we do want to reduce the log noise (#16), but I wasn't comfortable just going ahead and deleting these. Will wait for feedback from @flekschas before merging.

.eslintrc
```diff
@@ -36,6 +36,7 @@
     "guard-for-in": 0,
     "jsx-quotes": 1,
     "no-bitwise": 0,
+    "no-console": 0,
```
flekschas: This seems _dangerous_ to me. I'd at least go for warning otherwise we might accidentally commit logging statements. One can also conditionally allow `.warn()` and `.error()` logs if that's what we want.

mccalluc: Thanks: I've changed it so we only allow `console.warn()` and `console.error()`, but `log` and `info` will not be allowed. Longer term, we want a logging framework, but this is good enough for now?

PR #361, opened 2018-11-06
Do we need `npm run compile` twice?

mccalluc:
(I'll default to listing you both as reviewers going forward, and will generally wait for a green check mark from Pete, and no objections from Fritz. Let me know if some other approach would be better.)


flekschas:
> Do we need `npm run compile` twice?
Can you point out where the first compile is happening?

Without compiling, where is `dist.zip` coming from?

pkerpedjiev:
I second Fritz's question.

mccalluc:
@pkerpedjiev @flekschas : There had been two in travis.yml, and one is now in test.sh. I just added a check there to make sure dist.zip exists afterwards. I don't think running the tests would remove it, so I don't think it needs to be generated again.

mccalluc:
Renamed; thanks! Give me a green checkmark, and I'll merge it.

travis_build_test.sh
```diff
@@ -13,6 +13,7 @@ end eslint
 
 start compile
 npm run compile
+[ -e dist.zip ] || die 'Missing dist.zip'
```
pkerpedjiev: Maybe say `Missing dist.zip, please check that it was produced by npm compile in build.sh`

flekschas: I 😍
@mccalluc other than that please merge

Issue #360, opened 2018-11-06
Production build is broken


flekschas:
_Commit_: Since v1.2.4

#### Steps to reproduce
1. Use `https://unpkg.com/higlass@1.2.8/dist/hglib.min.js`

#### Observed behavior
- `window.hglib not defined`

mccalluc:
I get this console error: 
```
Uncaught TypeError: Cannot read property 'rgb2hex' of undefined
```
I'll try to see what commit caused the problem and propose a fix.

flekschas:
Ah but good pointer! I forgot to link the two!

mccalluc:
@flekschas : I believe this is fixed? Or if you'd prefer to keep this open till the next release, reopen?

PR #362, opened 2018-11-06
Flekschas/fix update build

flekschas:
- Fixed the minified build.
- White-listed only the dist files for faster `npm install`
- Cleaned up the `package.json` `scripts`
- Removed unused `page.css`
- Updated some webpack related libraries


flekschas:
This PR should fix #360

PR #363, opened 2018-11-06
Mccalluc/fix export heatmap scale

mccalluc:
Color scale is now correct on export. While I was here, linted the file.

(The exported scale still has extra tick marks at the bottom, and an extra vertical line.
screen shot 2018-11-06 at 4 18 15 pm

I've just been testing manually, but I should write a unit test. Also wondering what other cases should be checked... non-logarithmic? others?)


mccalluc:
huh: `Executed 207 of 231` and then it stalls. Let me try locally, and I may push some logging statements to travis if I'm still confused.

pkerpedjiev:
I think a test for linear scales and then zooming to another location and checking again should be OK to test this behavior.

Otherwise it looks fine.

mccalluc:
One of the tests is timing out, both local and remote:
```
09 11 2018 17:28:16.664:INFO [Chrome 70.0.3538 (Mac OS X 10.11.6) | Simple HiGlassComponent, Options tests | creates an editable component]: Success: 237 ms
09 11 2018 17:28:26.667:WARN [Chrome 70.0.3538 (Mac OS X 10.11.6)]: Disconnected (1 times), because no message in 10000 ms.
09 11 2018 17:28:26.667:ERROR [Chrome 70.0.3538 (Mac OS X 10.11.6)]: Disconnected, because no message in 10000 ms.
```
I'll track that down. Sorry.

pkerpedjiev:
If you think the test coverage is sufficient, then just go ahead and merge it.

mccalluc:
@pkerpedjiev or @flekschas : So there seems to be something in the `PngExportTest.js` that stalls: In travis there are no failing tests, but only 207 tests are run, but disable that one test and everything runs. I have not been able to reproduce this reliably locally.
- Have you had trouble with `PngExportTest`?
- Do you see anything in this PR that would be likely to cause problems with it, if it wasn't already having problems?
- Thoughts on the best way forward?

pkerpedjiev:
I think I fixed it. It was because `waitForTilesLoaded` was before `waitForJsonLoaded` in `mountHGComponent`. The viewconf itself is a JSON so we need to wait to fetch that before we can wait for the tiles.

pkerpedjiev:
If there's nothing else, feel free to merge this.

PR #367, opened 2018-11-07
Flekschas/add label margin

flekschas:
Added 4 new options to enable applying margin to labels:

```javascript
{
  labelLeftMargin: 10,  // applied on label positions topLeft and bottomLeft
  labelRightMargin: 10,  // applied on label positions topRight and bottomRight
  labelTopMargin: 10,  // applied on label positions topLeft and topRight
  labelBottomMargin: 10,  // applied on label positions bottomLeft and bottomRight
}
```

**Example**

screen shot 2018-11-07 at 4 36 49 pm

(See `bar-track-color-range.html` for a live demo)

**Update**

PR needs:

- [x] Add new cool feature (add margin)
- [x] Test margins
- [x] Document margins
- [x] Make margins SVG exportable


pkerpedjiev:
Without your comments it's hard to know where the labels are being moved. How about something like `marginToLabelPositionXAxis` and `marginToLabelPositionYAxis` or something like that?

flekschas:
My naming is inspired by CSS. If you have a `div` and set `margin-top:10px` then it's pushed down 10px from it's original y position. This is exactly what's happening with the labels. The only difference is that _opposing_ margins are ignore. In CSS one could push an element upwards by specifying `margin-bottom` to be larger than the height of the elements contain minus it's own height. This is not supported right now since the same effect can be achieved with negative top margin (e.g., `labelTopMargin: -10` moved the label 10px upwards) I wouldn't know why anyone would ever want to do this but we could implement it in the future.

To me this is more intuitive than thinking about a continuous X axis. Is that's more a Python thing?

pkerpedjiev:
Ok, that makes sense. Is there documentation for that? And tests?

flekschas:
3x no. 😅
I will add docs and see if I can extend an existing test for the label pos (does it exist?). I have no idea how the SVG renderer works. Is doing something automatic or essentially a re-implementation of the same PIXI.js logic?

pkerpedjiev:
> I can extend an existing test for the label pos (does it exist?)

I quickly checked in `HiGlassComponentTest` and didn't find anything. Just make a new file.

> Is doing something automatic or essentially a re-implementation of the same PIXI.js logic?

Unfortunately mostly the latter. Some tracks are clever and store locations of rectangles and colors, but most just have a reimplementation of the rendering function. The label drawing is handled in `PixiTrack.js` and most of it is implemented. You just have to modify it slightly to work with your new options.

I find it helpful to set up a test harness that automatically exports to SVG so that I can work on the implementation and see the output without constantly having to use the mouse to click on export to SVG. After that I just add a quick check or two to make sure the SVG output makes some sense. Even just grepping for a string is useful.

flekschas:
@Pete Where are we documenting globale track options? I might just not find it but where did you document things like `labelPosition`, `labelColor`, etc? Picking an arbitrary track doesn't make sense to me.

pete:
@flekschas I appreciate the request for my input, but I cannot say where such a thing is documented, being wholly unfamiliar with this codebase and the patch in question.

pete:
@flekschas No worries, I get random notifications a lot.  My hobby is to always reply to them as sincerely as possible.

I think they do autocomplete basically anyone.

pkerpedjiev:
Either here: https://github.com/higlass/higlass/blob/develop/docs/view_config.rst

or...

In the track types section here: https://github.com/higlass/higlass/blob/develop/docs/track_types.rst

Probably better in the latter.

PR #368, opened 2018-11-07
Mccalluc/overlays again

mccalluc:
- Would we ever worry about having too many examples, or is it fine to add this?
- Are there any particular standards for the examples? At some point I feel like having some documentation in there about what makes it interesting / useful could be good.
- Chromosome2DAnnotation feels anomalous: My sense was that HiGlass generally uses absolute coordinates, but this seems to be an exception.
- I looked at Annotation2DTrack: It seems like there should be something with an interface closer to the 1DTrack, but there aren't any examples of it in use.
- I would like to use promises the right way, rather than just chaining all these callbacks.
- Still need to add the interactivity.
- I could get more realistic data for the 2d.


mccalluc:
One build just stalled:
```
npm WARN deprecated browserslist@2.11.3: Browserslist 2 could fail on reading Browserslist >3.0 config used in other tools.
npm WARN deprecated browserslist@1.7.7: Browserslist 2 could fail on reading Browserslist >3.0 config used in other tools.
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```

pkerpedjiev:
> Would we ever worry about having too many examples, or is it fine to add this?

One can never have too many examples!

> Are there any particular standards for the examples? At some point I feel like having some documentation in there about what makes it interesting / useful could be good.

Not currently. You're welcome to create some though. At some point it would be nice to combine these with the ones here: http://test2.higlass.io/examples. Those are currently defined by this gist: https://gist.github.com/pkerpedjiev/104f6c37fbfd0d7d41c73a06010a3b7e All of those examples link to viewconfs and can thus be opened by higlass.io/app. The ones in this repository can be more complex because they may include API calls or animations or event handlers so I'm not sure if it's possible to combine the two.

At the very least, it would be nice to have an index listing the available examples, if that doesn't exist yet.

> Chromosome2DAnnotation feels anomalous: My sense was that HiGlass generally uses absolute coordinates, but this seems to be an exception.

Agreed. That's what that one Slack thread was about. It's used by HiPiler though. With that said, it would be convenient for the user to list regions using chromosome coordinates but I feel like that should be built on top of a generic absolute coordinate implementation and take a set of chromosome sizes as in its viewconf definition.

Your other comments seemed more like remarks for future notice. Let me know if you there's anything concrete there that I can help with.

pkerpedjiev:
Should I merge this PR or was there more that you wanted to do with it?

mccalluc:
If this much looks good to you, I'll merge it, and continue. Thanks!

PR #371, opened 2018-11-09
lint all the tests

mccalluc:
By lines, the fixes to the json blobs are the largest, and maybe it would be better to use ignores rather than fixing all those lines? Might be easier to review commits rather than the whole diff. The rest is pretty straight-forward, except for HiGlassComponentTest, which eslint can't even parse? Ignore now and if this is merged, I'll file a bug to remind me to come back to it.


flekschas:
Can be merged from my side after conflicts are resolved. Besides fixing things manually we should really set up precommit linting and auto code style fixing. I started using `eslint` + `husky` + `lint-staged` + `prettier` on another React and it works nicely. Of course this is would be a different PR :) Just throwing it out here.

mccalluc:
@pkerpedjiev : Does this look ok to you?

@flekschas : Thanks; just merged and resolved conflict. I have mixed feelings about pre-commit hooks. I wouldn't object to a PR, but
- You still need to run the checks in CI, in case someone hasn't installed the hooks.
- While working on a branch, I often push to github as my backup strategy, and I don't think every incremental commit needs to lint perfectly, just as not every incremental commit needs to pass all tests.
- It shouldn't happen, but as linting is done in more environments, things can get out of sync: On Refinery, the pre-commit linting and the CI-linting disagree on some things, and we don't understand why, exactly.

test/AddAndRemoveViewconfTests.js
```diff
@@ -34,8 +34,6 @@ describe('Simple HiGlassComponent', () => {
 
       api = viewer(div, simpleCenterViewConfig, {});
 
-      const p = api.setViewConfig(simpleCenterViewConfig);
```
pkerpedjiev: Are you removing this because the return value isn't used? 

It might be there just to be called and ensure no errors are raised.

test/BedLikeTests.js
```diff
@@ -22,6 +22,73 @@ describe('Simple HiGlassComponent', () => {
   let hgc = null;
   let div = null;
 
+  const viewConf = {
```
pkerpedjiev: This is one thing that annoys me about linting. I had intentionally put that viewconf at the bottom because what's important is the test. Now that it's at the top, we have to scroll past it to get to the interesting test. This lengthy viewconf is now obscuring the test to conform to a style guide that is supposed to make the code more legible. Ironic.

Maybe we can disable that rule? What do you think?

test/HiGlassComponentTest.js
```diff
@@ -672,9 +672,7 @@ describe('Simple HiGlassComponent', () => {
       const track = getTrackByUid(views.aa.tracks, 'heatmap1');
       track.options.colorbarPosition = 'hidden';
 
-      hgc.instance().setState(
```
pkerpedjiev: How did this even run? It looks like it used to be syntactically incorrect.

test/AddAndRemoveViewconfTests.js
mccalluc: Sorry: That was too aggressive. I'll put it back, but just won't assign it to anything.

pkerpedjiev:
I have a feeling that this going to cause a whole mountain of conflicts when merged with Fritz's pubSub changes. Somebody else can hit the green button 🙈

flekschas:
@pkerpedjiev the merge was surprisingly simple! :)

PR #372, opened 2018-11-09
Re-add example

flekschas:
Initially added with https://github.com/higlass/higlass/pull/285

@mccalluc This track might be of interest to you

screen shot 2018-11-09 at 1 14 26 pm

Docs are here: https://github.com/higlass/higlass-docs-dev/blob/master/track_types.rst#1d-annotations

### DIFF
--- a/docs/examples/1d-annotations.html
+++ b/docs/examples/1d-annotations.html
@@ -1,0 +1,382 @@
+<!DOCTYPE html>
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width,initial-scale=1">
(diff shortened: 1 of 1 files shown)


pkerpedjiev:
Nice job on having docs for this track! http://docs.higlass.io/track_types.html#d-annotations

Is stroke-pos not settable through the UI on purpose?

image

flekschas:
I guess not. I'll create a ticket for it. Do you mind if we keep that to another PR? This one is really just about adding the lost example again.

pkerpedjiev:
> Do you mind if we keep that to another PR?

Nope, that's fine.

PR #375, opened 2018-11-09
Flekschas/add interaction handling to 1d annotations

flekschas:
Making 1D annotations clickable and the click event capturable through the JS API like so:

```
hgApi.on('click', ({ type, event, payload }) => {
  console.log('Click', type, event.data.originalEvent);
});
// --> Click annotation (2) [0, 100] PointerEvent {isTrusted: false, sourceUid: "EQKOr4AHSFaQOaSoiNDQuw", forwarded: undefined, pointerId: 1, width: 1, …}
```

Right now only 1D annotations capture click events (note that PIXI is not able to actually capture click events with our current setup but is able to capture _mousedown_ and _mouseup_ events)

Example: Added (`docs/1d-annotations.html`)
Docs: Updated (`docs/higlass_developer.rst`)
Test: Passing (`test/test/APITests.js`)

- [x] Adds interactions
- [x] Test interactions
- [x] Documents interactions
- [x] Provides example


app/scripts/Annotations1dTrack.js
```diff
@@ -3,21 +3,23 @@ import * as PIXI from 'pixi.js';
 
 import PixiTrack from './PixiTrack';
 
+import pubSub from './services/pub-sub';
+
+const MAX_CLICK_DELAY = 300;
```
pkerpedjiev: What's this for?

app/scripts/Annotations1dTrack.js
```diff
@@ -53,10 +55,21 @@ class Annotations1dTrack extends PixiTrack {
     const graphics = this.pMain;
     graphics.clear();
 
+    const timeStamp = performance.now();
```
pkerpedjiev: How is this used?

app/scripts/Annotations1dTrack.js
```diff
@@ -164,13 +177,44 @@ class Annotations1dTrack extends PixiTrack {
...
+          start, 0, width, this.dimensions[1]
+        );
       }
+
+      this.rects[id].graphics.mousedown = () => {
+        this.rects[id].mouseDownTime = performance.now();
+      };
+
+      this.rects[id].graphics.mouseup = (event) => {
+        if (performance.now() - this.rects[id].mouseDownTime < MAX_CLICK_DELAY) {
```
pkerpedjiev: Why is this necessary? Are you trying to distinguish from dragging?

app/scripts/HiGlassComponent.js
```diff
@@ -230,6 +230,7 @@ class HiGlassComponent extends React.Component {
     this.zooming = false;
 
     // Bounded functions
```
pkerpedjiev: I know this isn't part of the PR, but should this comment actually say "Bound functions"?

flekschas:
@pkerpedjiev I've copied over the docs to the docstring in `api.js` and added comments regarding your questions from above.

app/scripts/Annotations1dTrack.js
```diff
@@ -3,21 +3,28 @@ import * as PIXI from 'pixi.js';
 
 import PixiTrack from './PixiTrack';
 
+import pubSub from './services/pub-sub';
+
+// Maximum delay in ms between mousedown and mouseup that is registered as a
+// click. Note we need to use mousedown and mouseup as PIXI doesn't recognize
+// click events with out current setup. Since most UIs treat long clicks as
+// either something special or a cancelation we follow best practices and
```
mccalluc: Do you understand what it is that prevents click events from being registered? ... Sounds like it's something we do, and not something intrinsic to Pixi?

flekschas: Yes I know what it is and yes it's something on our side. If you want hover you need to constantly flush the PIXI object. That's fine for gaming but for a web app it's bad because it drains the battery but using too much CPU even if nothing changes. Why are you asking?

app/scripts/Annotations1dTrack.js
```diff
@@ -164,13 +186,45 @@ class Annotations1dTrack extends PixiTrack {
...
+          pubSub.publish('app.click', {
+            type: 'annotation',
+            event,
+            payload: region
+          });
+        }
+      };
     });
+
+    // Remove outdated rects, i.e., rects whos time stamp is not the current
```
mccalluc:
```suggestion
    // Remove outdated rects, i.e., rects whose time stamp is not the current
```

app/scripts/Chromosome2DAnnotations.js
```diff
@@ -4,11 +4,20 @@ import * as PIXI from 'pixi.js';
 import PixiTrack from './PixiTrack';
 import ChromosomeInfo from './ChromosomeInfo';
 
+import pubSub from './services/pub-sub';
+
+// Maximum delay in ms between mousedown and mouseup that is registered as a
+// click. Note we need to use mousedown and mouseup as PIXI doesn't recognize
+// click events with out current setup. Since most UIs treat long clicks as
```
mccalluc:
```suggestion
// click events with our current setup. Since most UIs treat long clicks as
```

test/APITests.js
```diff
@@ -120,11 +146,63 @@ describe('Simple HiGlassComponent', () => {
...
+        setTimeout(() => {
+          canvas.dispatchEvent(createPointerEvent('pointerdown', 100, 100));
+          canvas.dispatchEvent(createPointerEvent('pointerup', 100, 100));
+        }, 50);
+
+        setTimeout(() => {
+          expect(clicked).to.equal(2);
+          api.destroy();
+          done();
+        }, 100);
```
mccalluc: Might set all the timeouts to 0: They should be queued and just run one after the other, in the order they are registered. If waiting 50ms or 100ms actually is necessary that would be a race condition it would be good to know about.

app/scripts/Annotations1dTrack.js
mccalluc: Thanks: The context helps me. That was what I wanted to know... It's something we could turn on, but we have good reasons not to. In the original context it sounded like it might be a bug we didn't understand, or something else.

flekschas: If you're curious about the details, take a look here: https://stackoverflow.com/questions/27025843/interactive-graphics-with-pixijs-mouseover-doesnt-fire-while-click-does
The expensive thing (which is not a problem for games but for us) is looping updater which leads to a constant CPU usage of ~10-15% even when nothing changes. Please don't ask me why click events work for them but not us... I assume it's because of our event forwarding system but I could be wrong. 😅 I am just happy the simulated click works

app/scripts/Chromosome2DAnnotations.js
```diff
@@ -4,11 +4,20 @@ import * as PIXI from 'pixi.js';
...
+// either something special or a cancelation we follow best practices and
+// implement a threshold on the delay as well.
+const MAX_CLICK_DELAY = 300;
```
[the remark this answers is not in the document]
flekschas: > Since most UIs treat long clicks as either something special or a cancelation we follow best practices and implement a threshold on the delay as well.

In general you're right but then there's no way to cancel a click. Feel free to propose an alternative.

> Making the whole thing inherit from some base class might be hard, but could this value be stored in one place?

I'll move `MAX_CLICK_DELAY` to `config/index.js`

app/scripts/HiGlassComponent.js
```diff
@@ -3346,6 +3348,13 @@ class HiGlassComponent extends React.Component {
     select('body').selectAll('.track-mouseover-menu').remove();
   }
 
+  /**
+   * Handle internally broadcasted click events
+   */
+  appClickHandler(data) {
+    apiPublish('click', data);
```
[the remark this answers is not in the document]
flekschas: I am re-publishing the click on the API's pubSub. (Notice the naming difference `publish` vs `apiPublish`. In this case it might seem overkill but in general the app could do something in response to the click event before sending it out _into the world_ (i.e., the public API)

test/APITests.js
```diff
@@ -25,7 +27,31 @@ import {
...
 
+function findCanvas(element) {
+  if (element.tagName.toLowerCase() === 'canvas') return element;
+  let canvas;
+  element.childNodes.forEach((childElement) => {
+    const el = findCanvas(childElement);
+    if (el) canvas = el;
+  });
+  return canvas;
+}
```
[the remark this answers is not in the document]
flekschas: Good catch. Will fix.

PS: Your code won't work because you can't break out of `forEach`.

test/APITests.js
flekschas: Will test but the tests fail in general because there is / are more nasty issues. The HG instances or the tests are somehow non-independent and I have no idea how to fix those. Changing the order of API tests will make others fail while this one succeeds. Any input on that would be much appreciated.

test/APITests.js
```diff
@@ -25,7 +27,31 @@ import {
...
+
 function createElementAndAPI(viewConfig, options) {
+  // Tests do not seem to be independent. I tried to enforce that existing DOM
+  // children of `body` are removed (and react apps unmounted) but although this
+  // should not have any impact on the test it actually leads to weird errors
+  // indicating that there is some shared global state between HG instances.
```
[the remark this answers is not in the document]
flekschas: This might actually get fixed with #370 (at least to some extent)

docs/higlass_developer.rst
```diff
@@ -465,19 +465,31 @@ HiGlass exposes the following event, which one can subscribe to via this method:
...
-``location:`` Returns an object describing the visible region
+``click``: Returns clicked objects. (Currently only clicks on 1D annotations are captured.)
+
+.. code-block:: javascript
+
+    {
+      type: 'annotation',
+      event: { ... },
+      payload: [230000000, 561000000]
+    }
```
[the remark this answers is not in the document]
flekschas: There are two ways to achieve this similar to normal event listeners:

1. compare to your local reference object

   ```
   const locusA = [...];
   const locusB = [...];
   const loci = [locusA, locusB];
   
   ...
   options: { regions: loci }
   ...

   hgApi.on('click', ({ payload }) => { if (payload === locusA) console.log(🎉); }
   ```

2. specify an ID upon instantiation of HG

   ```
   const loci = [[..., 'locusA'], [..., 'locusB']]
   hgApi.on('click', ({ payload }) => { if (payload[9] === 'locusA') console.log(🎉); }
   ```

app/scripts/api.js
```diff
@@ -493,6 +493,10 @@ const api = function api(context) {
         : listenerId;
 
       switch (event) {
+        case 'click':
+          apiPubSub.unsubscribe('click', callback);
```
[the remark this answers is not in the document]
flekschas: that's what I am talking about! what?

app/scripts/Annotations1dTrack.js
flekschas: I though I replied but maybe I didn't.

This is used for distinguishing between a canceled click (long click) and a normal click.

app/scripts/Annotations1dTrack.js
flekschas: I though I replied but maybe I didn't.

This is used to figure out which rect graphics object are still visible and which ones can be destroyed. For every draw cycle a new time stamp is generated and all visible rects get this new time stamp. rect graphics without this current time stamp are then destroyed.

app/scripts/Annotations1dTrack.js
flekschas: I though I replied but maybe I didn't.

This is used for distinguishing between a canceled click (long click) and a normal click

app/scripts/Annotations1dTrack.js
pkerpedjiev:
```suggestion
// for distinguishing between a cancelled click (long click) and
// a normal click
const MAX_CLICK_DELAY = 300;
```
More comments, more better.

pkerpedjiev:
I'm approving this but @mccalluc should also take a look and ensure that he's OK with it.

docs/examples/1d-annotations.html
```diff
@@ -378,5 +378,11 @@
 window.hgApi = window.hglib.viewer(
   document.getElementById('demo'), viewConfig, { bounded: true },
 );
+
+window.hgApi.on(
+  'click', ({ type, event, payload }) => {
+    console.log('Clicked on ', type, '\n', event, '\n', payload);
```
mccalluc: This might be the one case where I'd prefer an alert, or else add text to explain that the console needs to be opened to see the result of the event. (But I think we generally could do better at explaining what our demos are supposed to demonstrate, and this is just one instance.)

flekschas: I agree but it depends on the purpose of the examples. I started those simply to have a place for developers to see how certain things can be used. Anyone who knows what a click event listener is should be able to look at the code and understand what's going on. The concept is similar to http://regl.party/examples where the code is simultaneously shown next to the visual output. Something I think would be awesome to have in general.

PR #377, opened 2018-11-10
Pull docs from higlass-docs-dev rather than higlass-docs

pkerpedjiev:
This just pulled `higlass-docs-dev:master` as a subtree into `docs`.


pkerpedjiev:
>  if the automatic doc builds seems to be working, we can get rid of doc repos entirely?

Yes.

PR #387, opened 2018-11-13
Pkerpedjiev/nested checkboxes

### LINKED ISSUE #384: Add nesting to the tileset list
The current tileset list is overwhelming:

image

There lots of datasets listed without any grouping. We need a way to partition them into groups so that the list can be cleaned up.

## Steps to reproduce

1. Go to higlass.io/app
2. Add dataset -> center
3. See tileset list

One option is to use [react-checkbox-tree](https://github.com/jakezatecky/react-checkbox-tree).

pkerpedjiev:
## Description

This change introduces nested checkboxes to the tileset finder (Add Track) dialog. This makes it possible to assign a group to each tileset and to group them by their groups. 

image

This should clean up the list of tilesets making it easier for users to find what they want.

Fixes #384
Goes with (requires): https://github.com/higlass/higlass-server/pull/78

## Checklist

- [NA] Example added or updated


pkerpedjiev:
Waiting on https://github.com/higlass/higlass/pull/427 so that I can update the docs.

app/scripts/AddTrackModal.js
```diff
@@ -148,23 +156,25 @@ class AddTrackModal extends React.Component {
...
           { form }
           {
-            this.props.hidePlotTypeChooser ? null :
-            <PlotTypeChooser
-              ref={(c) => { this.plotTypeChooser = c; }}
-              datatypes={this.state.selectedTilesets.map(x => x.datatype)}
-              onPlotTypeSelected={this.handlePlotTypeSelected.bind(this)}
-              orientation={orientation}
-            />
+            this.props.hidePlotTypeChooser ? null
```
flekschas: Not critical but this can be shortened to

```javasacript
!this.props.hidePlotTypeChooser && (...)
```

app/scripts/PlotTypeChooser.js
```diff
@@ -91,7 +91,13 @@ class PlotTypeChooser extends React.Component {
     }
 
     return (<div>
-      { AVAILABLE_TRACK_TYPESList }
+      { AVAILABLE_TRACK_TYPESList.length ?
```
flekschas: Same as above instead of `?` one can use `&&` to not having to deal with `: null`. Again this is not important, just FYI

app/scripts/TilesetFinder.js
```diff
@@ -188,7 +176,7 @@ class TilesetFinder extends React.Component {
   }
 
   handleSelect() {
-    const { selectedOptions } = ReactDOM.findDOMNode(this.multiSelect);
+    const { selectedOptions } = this.multiSelect;
```
flekschas: Just double checking: `ReactDOM.findDOMNode` is no longer needed?

app/scripts/TilesetFinder.js
```diff
@@ -200,70 +188,142 @@ class TilesetFinder extends React.Component {
...
   }
 
+  /*
+   * Create the nested checkbox tree by partitioning the
+   * list of available datasets according to their group
+   *
+   * @param {dict} datasetsDict A dictionary of id -> tileset
+   *  def mappings
+   * @param {string} filter A string to filter the results by
+   * @returns {list} A list of items that define the nested checkbox
```
flekschas: I'd use proper JS dtypes: `dict` => `object` and `list` => `array`

app/scripts/TilesetFinder.js
```diff
@@ -200,70 +188,142 @@ class TilesetFinder extends React.Component {
...
+   *  structure
+   *  {
+   *    'name': 'blah',
+   *    'value': 'blah',
+   *    'children': 'blah',
```
flekschas: Looking at line `216` it seems that children is suppose to be an `array` and not a string? Is that right or am I misunderstanding something?

app/styles/TilesetFinder.css
```diff
@@ -0,0 +1,249 @@
+@font-face {
+  font-family: 'FontAwesome';
+  src: url('https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/fonts/fontawesome-webfont.eot?v=4.7.0');
```
flekschas: What do we need fontawesome for? Could we just download and inline the icon? External references can lead to issues when being offline or within some secret organizations intranet.

karma.conf.js
```diff
@@ -13,8 +13,10 @@ module.exports = (config) => {
       'node_modules/pixi.js/dist/pixi.js',
       'node_modules/react-bootstrap/dist/react-bootstrap.js',
       'node_modules/bootstrap/dist/css/bootstrap.min.css',
+      'node_modules/font-awesome/css/font-awesome.css',
```
flekschas: If possible I would be great if we can avoid adding another big dependency :)

package.json
```diff
@@ -113,6 +114,7 @@
     "eslint-plugin-react": "^7.11.1",
     "fast-css-loader": "^1.0.2",
     "fast-sass-loader": "^1.4.6",
+    "font-awesome": "^4.7.0",
```
flekschas: As above. Just pinning

webpack.config.js
```diff
@@ -32,6 +32,7 @@ module.exports = (env, argv) => ({
       path.resolve(__dirname, 'app'),
       path.resolve(__dirname, 'docs', 'examples'),
       path.resolve(__dirname, 'node_modules'),
+      path.resolve(__dirname, 'lib', 'vendor')
```
flekschas: What's this used for?

app/scripts/AddTrackModal.js
pkerpedjiev: Not in this case. Notice that it's `a ? null : b`, which isn't the same as `a && b`.

app/scripts/PlotTypeChooser.js
pkerpedjiev: Here it works.

app/scripts/TilesetFinder.js
pkerpedjiev: Sure.

app/scripts/TilesetFinder.js
pkerpedjiev: Nope.

app/scripts/TilesetFinder.js
pkerpedjiev: No, my docstring was just wrong. Good catch.

app/styles/TilesetFinder.css
pkerpedjiev: If some secret organizations' internet isn't connected they're going to have bigger fish to fry than the icons here.

But nevertheless, I included them as inlined SVGs.

karma.conf.js
pkerpedjiev: Ok, it's gone.

package.json
pkerpedjiev: Gone.

webpack.config.js
pkerpedjiev: I just added it to put local libraries so that I could work on HiGlass while on the plane :)

flekschas:
Awesome! Thanks!

The PR TravisCI check fails but it seems like it's an issue on Travis side

PR #392, opened 2018-11-13
Mccalluc/better use of promises in finn demo

mccalluc:
- Before, all the outside files were loaded synchronously. Now, just the chrom info needs to be loaded first, and then the CSVs can be loaded in parallel.
- Added a blurb up top to explain what this example is about.
- Simplified some of the framing CSS/HTML, but not sure this is the best way of doing it.

### DIFF
--- a/docs/examples/finn-et-al/index.html
+++ b/docs/examples/finn-et-al/index.html
@@ -30,9 +30,8 @@
 <script src='makeViewConf.js'></script>
 <script>
 (() => {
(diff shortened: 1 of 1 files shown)


pkerpedjiev:
But, having not seen the actual example, why are the 1D annotations being converted to absolute coordinates and the 2D annotations not being converted?

mccalluc:
The 2d track type is chromosome aware... I think that was part of the long thread between you and Fritz? I think he's planning to make it (or probably a new type, actually) work with absolute coordinates. 

As far as documentation, we might want some notion of track types which are deprecated, either because their interface is non-standard, or because they were never completely implemented.

PR #407, opened 2018-11-15
Create a separate datastructure which both renderers use

### LINKED ISSUE #345: Handle clipping + NANs consistently in Canvas and SVG
_Commit_: 98576dd

#### Steps to reproduce
1. Add a line track
2. Zoom until there are breaks
3. Export SVG

#### Observed behavior
screen shot 2018-10-30 at 10 27 42 am

#### Expected behavior
Rendering should be the same in either mode, but I think that neither of these is actually correct. We should not draw any line segments to the top. When there is just a single point, with its neighbors on either side being NAN, I'm not sure what the best behavior is, but perhaps draw nothing for a first iteration? Usually fetching a deeper tile can provide some context around an isolated point, but that's complicated.

mccalluc:
## Description

- We first create a list of segments, and then both renders use the same data structure for drawing.

Why is it necessary?
- Previously, when the log was taken of 0, the chart would appear to be clipped, with different behavior on the rising and falling slopes. It is now consistent and correct between renderers

Fixes #345

## Checklist

- [ ] Unit tests added or updated ... but the easiest way to test this would be a unit test just around the segment data structure... but to do that, I'd want to pull out these changes and put them in a helper function. Would that be good? 

screen shot 2018-11-15 at 5 06 05 pm

### DIFF
--- a/app/scripts/HorizontalLine1DPixiTrack.js
+++ b/app/scripts/HorizontalLine1DPixiTrack.js
@@ -162,32 +162,39 @@ class HorizontalLine1DPixiTrack extends HorizontalTiled1DPixiTrack {
     const strokeWidth = this.options.lineStrokeWidth ? this.options.lineStrokeWidth : 1;
     graphics.lineStyle(strokeWidth, stroke, 1);
 
-    const logScaling = this.options.valueScaling === 'log';
+    tile.segments = [];
+    let currentSegment = [];
 
     for (let i = 0; i < tileValues.length; i++) {
       const xPos = this._xScale(tileXScale(i));
       const yPos = this.valueScale(tileValues[i] + offsetValue);
 
-      tile.xValues[i] = xPos;
-      tile.yValues[i] = yPos;
-
-      if (i === 0) {
-        graphics.moveTo(xPos, yPos);
+      if (this.options.valueScaling === 'log' && tileValues[i] === 0) {
+        if (currentSegment.length > 1) {
+          tile.segments.push(currentSegment);
+        }
+        // Just ignore 1-element segments.
+        currentSegment = [];
         continue;
       }
 
       if (tileXScale(i) > this.tilesetInfo.max_pos[0]) {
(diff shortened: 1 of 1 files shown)


flekschas:
I am not sure what this PR is about. Can you clarify the following:

- What am I suppose to see in the screenshot? Which track is new? What exactly has changed except for the shadow, the blue axis and the labels, or is that what you want to point out?
- What is this `list of segments`? Is this list of segments new? Was it there before?

Maybe I am not the right person to review this.

mccalluc:
Sorry not to provide better context. The point of the screenshot is that the lines are the same, particularly where there are gaps in the data. They should be compared with the screenshots from the original bug:
screen shot 2018-10-30 at 10 27 42 am

The segements is just a list of pairs that I build. Previously, the drawing directives were in the loops that also had the special case handling for out-of-range values, and we were handling them differently between pixi and svg.

pkerpedjiev:
I think the idea is to store the regions that are drawn in a datastructure rather than iterating over the data in both SVG and PIXI renderers. I think this is a good idea and there's precedent for it in other tracks where we store `drawnRects` or in the multivec track where we store locations of the rectangles to be drawn and then use those to draw both to PIXI and to SVG.

Apart from the axes, I'm also having a hard time seeing a difference between the two screenshots.

> I'd want to pull out these changes and put them in a helper function. Would that be good?

Sounds fine to me. 

As a side note, be aware that there's a gap between tiles. If you're working in this area you might have ideas for how to fix this.

mccalluc:
Sorry: I'll try to explain the screenshots again: 

Before: SVG and Canvas *are different*, though they both are clipped in areas where we're trying to take log 0.
screen shot 2018-10-30 at 10 27 42 am

After: SVG and Canvas *are the same*. (Axes are not part of this PR.) Where we try to take log 0 there is simply a gap.
screen shot 2018-11-15 at 5 06 05 pm

PR #410, opened 2018-11-16
SVG Axis was missing bits from Pixi

### LINKED ISSUE #344: Inaccurate heatmap colorscale in SVG export
_Commit_: 98576dd

#### Steps to reproduce
1. Load the development demo
2. To make it more noticeable, drag the handles at the top and bottom of the colorscale for the heatmap to make a smaller range
3. Export to SVG

#### Observed behavior
screen shot 2018-10-30 at 10 43 57 am

#### Expected behavior
- The heatmaps are colored the same, but the exported scale is inaccurate.
- Also, make the ticks on the scale black.

mccalluc:
## Description

- Heatmap scales in SVG export have the correct scale, and color defaults to black.

Fixes #344 

Related to PR #363, but that only handled problems relating to the colors drawn, not to the axis numbers and ticks.

## Checklist

- [ ] Unit tests added or updated: none, but the real problem here is the copy-and-pasting of code, and I think cleaning up that would be the better investment of time. Thoughts?

left: before / right: after
top: pixi / bottom: svg

screen shot 2018-11-15 at 9 08 06 pm

### DIFF
--- a/app/scripts/AxisPixi.js
+++ b/app/scripts/AxisPixi.js
@@ -221,12 +221,13 @@ class AxisPixi {
 
     let stroke = 'black';
 
-    if (this.track) {
-      stroke = this.track.options.lineStrokeColor
-        ? this.track.options.lineStrokeColor
-        : 'blue';
+    if (this.track && this.track.options.lineStrokeColor) {
+      stroke = this.track.options.lineStrokeColor;
     }
-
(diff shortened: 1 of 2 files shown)


mccalluc:
Thanks! I'll hold off on the merge until `mccalluc/svg-preview` is approved, to avoid confusing things there.

PR #414, opened 2018-11-16
Restore working index.html

### LINKED ISSUE #413: Index is replaced by specific example and logs out dozens of errors
_Commit_: Related to #412

#### Steps to reproduce
1. Open `localhost:8080`
2. Open the web console

#### Observed behavior
```
96Error: <rect> attribute height: A negative value is not valid. ("-0.03087100330760606")
6Error: <rect> attribute y: Expected length, "NaN".
6Error: <rect> attribute height: Expected length, "NaN".
(index):46 Error: <path> attribute d: Expected number, "…1.9518868899945 NaNL883.17990878…".
setTimeout @ (index):46
setTimeout (async)
renderPreview @ (index):45
d3.json.then @ (index):50
Promise.then (async)
(anonymous) @ (index):37
(index):46 Error: <path> attribute d: Expected number, "…0.9518868899947 NaNL722.17990878…".
setTimeout @ (index):46
setTimeout (async)
renderPreview @ (index):45
d3.json.then @ (index):50
Promise.then (async)
(anonymous) @ (index):37
(index):46 Error: <path> attribute d: Expected number, "…1.9518868899945 NaNL883.17990878…".
setTimeout @ (index):46
setTimeout (async)
renderPreview @ (index):45
d3.json.then @ (index):50
Promise.then (async)
(anonymous) @ (index):37
```
- Local view confs are replaced by an external viewconfig
- d3 is included

#### Expected behavior
- The index should be error free
- The index should be somewhat neutral and not represent one specific use case
- Unnecessary libraries should be avoided

flekschas:
## Description

Fixes #413

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -9,16 +9,11 @@
 <style type="text/css">
 #demo {
   position: absolute;
-  left: 0;
-  top: 0;
-  bottom: 0;
-  right: 0;
+  left: 1rem;
+  top: 1rem;
+  bottom: 1rem;
+  right: 1rem;
   overflow: hidden;
-  height: 50%;
-}
-#preview {
-  position: absolute;
-  top: 50%;
 }
 </style>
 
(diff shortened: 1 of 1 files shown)


pkerpedjiev:
The full view + svg *should* be in the repo. We (I) often neglect SVG because it's not immediately visible. Having it front and center makes it harder to avoid. 

When I'm working on higlass I usually have a separate html file (test2.html) that has a viewconf particular to the problem I'm working. This file is in the .gitignore because it's just for development purposes. Having the template @mccalluc created is really useful for trying out new tasks.

I'm fine with keeping `index.html` as it was before and placing the @mccalluc's example in `svg.html`. I would also be in favor of adding a note to the `Development` section of the README indicating that `svg.html` exists and that it should be used when developing or modifying new tracks.

mccalluc:
Just merged #416, so SVG preview is now with the rest of examples, and I added a note to the README. I'll go ahead and merge this, now.

PR #416, opened 2018-11-17
Mccalluc/index tweaks

mccalluc:
## Description

- Move index.html to svg.html so index can be reverted.
- Add event handlers.
- Replace higlass.io html with higlass.io viewconf


app/svg.html
```diff
@@ -0,0 +1,63 @@
...
+<script src="https://cdnjs.cloudflare.com/ajax/libs/react-bootstrap/0.32.1/react-bootstrap.min.js"></script>
+<script src="http://d3js.org/d3.v5.min.js"></script>
+</head>
+<body>
+  <div id="demo"></div>
+  <div id="preview"></div>
+</body>
+<script src='hglib.js'></script>
+<script>
+d3.json(location.search.slice(1).replace('http://higlass.io/app/?config=', 'http://higlass.io/api/v1/viewconfs/?d=') || 'viewConf.json').then((viewConfig) => {
```
pkerpedjiev: Sorry, what is the `location.search.slice(1).replace` statement doing?

Also, just out of curiosity, why `d3.json` over `fetch`?

pkerpedjiev:
Approved. 

How do you feel about adding a note to the Development section of the README indicating that svg.html exists and that it should be used when developing or modifying new tracks?

app/svg.html
mccalluc: > Sorry, what is the location.search.slice(1).replace statement doing

The idea is that you can do this:
```
http://localhost:8080/svg.html?http://higlass.io/app/?config=MSHhOBbOSW6iIovB5yk6BA
```
and it will translate that into the url for the json, and view that. (Or you can reference a viewconf you've saved locally, or you can leave the query blank, and get the default.)

Maybe a little input box on the page that would prompt for a url to make this functionality more obvious?

> Also, just out of curiosity, why d3.json over fetch?

You mean the built in Fetch API? No good reason. I'll switch that over.

mccalluc:
I would like to move this over to the examples directory, rather than littering the top level.

>How do you feel about adding a note to the Development section of the README indicating that svg.html exists and that it should be used when developing or modifying new tracks?

Will do.

PR #419, opened 2018-11-20
Mccalluc/fix console error on default

mccalluc:
## Description

This gets rid of the console errors we had when trying to render the default svg.
(It's possible that the invalid values should not have gotten even this far, but this much feels like an improvement.)

I do not see any visual changes.

Also, in the SVG example, log the viewconf as an object, to keep the logs tidier.

## Checklist


pkerpedjiev:
Uh oh, I approved but the build is failing.

mccalluc:
Replaced by #438.

PR #423, opened 2018-11-21
Fix duplicated wheel forwarding

### LINKED ISSUE #417: Zoom behavior got much more sensitive since v1.2.8
_Commit_: develop

#### Steps to reproduce
1. zoom in and out

#### Observed behavior
- The zooming is much more sensitive, i.e., you zoom in and out way farther compared to the master branch.

In this example I am turning the mouse wheel about the same amount in `v1.2.8` (left window) and in `develop` (right window) but I am zooming in and out way farther on the right.

nov-18-2018 15-12-02

#### Expected behavior
- Zoom sensitivity should be the same as in the master branch (i.e., `v1.2.8`)

My best guess is that this is related to the event forwarding. In #370 I noticed that all events are triplicated. If this is true than it would mean that the zooming got 3x more sensitive. @pkerpedjiev do you by chance know if we change anything in `develop` related to the forwarding since `v1.2.8`?

flekschas:
The key aspect is in `TrackRenderer` line 1903. Setting `e.forwarded = true;` in the `TrackRenderer`ensures that the `wheelHandler` in `HiGlassComponent` doesn't fire again when the wheel event from `track-renderer-event` bubbled all the way up to the `window` which the `domEvent` service is listening to.

## Description

Remove duplicated wheel events

Why is it necessary?

Fixes #417 

## Checklist

### DIFF
--- a/app/scripts/HiGlassComponent.js
+++ b/app/scripts/HiGlassComponent.js
@@ -233,8 +233,6 @@ class HiGlassComponent extends React.Component {
     this.keyDownHandlerBound = this.keyDownHandler.bind(this);
     this.keyUpHandlerBound = this.keyUpHandler.bind(this);
     this.resizeHandlerBound = this.resizeHandler.bind(this);
-    this.mousewheelHandlerBound = this.mousewheelHandler.bind(this);
-    this.wheelHandlerBound = this.wheelHandler.bind(this);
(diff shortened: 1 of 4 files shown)


flekschas:
@pkerpedjiev Yes it does but better double check if you have the chance. Not that there are edge cases I forgot to test manually.

PR #427, opened 2018-11-23
Introduce the `fileUrl` tileset source 

pkerpedjiev:
## Description

This pull request has two parts:

1. Introduce a new way of retrieving data using `fileUrl` in the track viewconf. This only works if the specified `server` has the `/api/v1/register_url` endpoint. If it does, the one need not ingest an http-accessible file on the server. The register_url endpoint simply registers that url with the server and returns a tilesetUid. Tiles requests will be routed through the tileset server.

2. Documentation updates. Fixed some sphinx warnings, added requirements, added section about the `fileUrl` feature.

Why is it necessary?

This greatly simplifies the process of creating HiGlass-driven apps. Instead of having to perform an action on the server to add a new dataset, it can just be added to the viewconf. There is currently no UI element for selecting an url-based dataset but that can be included in another PR.

## Checklist

- [NA] Unit tests added or updated
    - We don't have a public working `fileUrl`-compatible server to run tests on yet. I have tested with a local implementation.
- [NA] Example added or updated
- [NA] Screenshot for visual changes (e.g. new tracks)


docs/higlass_developer.rst
```diff
@@ -1,699 +0,0 @@
-Developer
```
pkerpedjiev: All of this information is in `javascript_api.rst` so I removed the file.

pkerpedjiev:
Waiting on https://github.com/higlass/higlass/pull/427 so that I can update the docs.

.travis.yml
```diff
@@ -16,6 +16,10 @@ before_script:
 script:
     - ./travis_build_test.sh
 before_deploy:
+    - wget http://repo.continuum.io/miniconda/Miniconda3-4.3.21-Linux-x86_64.sh -O miniconda.sh
+    - bash miniconda.sh -b -p $HOME/miniconda
+    - export PATH="$HOME/miniconda/bin:$PATH"
+    - conda install --yes python=3.6
```
flekschas: Just out of curiosity, what do we need conda for on TravisCI?

app/scripts/DataFetcher.js
```diff
@@ -197,7 +257,7 @@ export default class DataFetcher {
     const result = new Float32Array(numeratorData.length);
 
     for (let i = 0; i < result.length; i++) {
-      if (denominatorData[i] === 0.) result[i] = NaN;
+      if (denominatorData[i] === 0.0) result[i] = NaN;
```
flekschas: FYI, I think `=== 0` and `=== 0.0` are equivalent because JS doesn't distinguish between `int` and `float` (sadly)

pkerpedjiev: Linter complains about 0.

.travis.yml
pkerpedjiev: Some package was complaining about an out-of-date python. Conda gives us a nice python3 environment where sphinx can install its dependencies.

pkerpedjiev:
Can we merge it now given that it passes all the other tests? It might be a while before we have working production test server to try it out with so it becomes a chicken and the egg problem.

As soon as the server is out, we can add a test viewconf to ensure that it functions as expected. In the meantime, it is marked as experimental in the documentation so hopefully nobody expects it to work perfectly.

pkerpedjiev:
Or, you know what? I'll separate out the documentation changes of this PR, put those in a PR and we can wait to merge this until later.

The server tests can be performed without the client but not the reverse. No chicken and egg problem :+1:

flekschas:
Either way is good with me but I guess v1.3+ makes sense if there isn't any server out yet either way

frankyan:
Hi, I want to know how to add one public higlass data (on 4DN data portal higlass server) into my local higlass server. It seems that the `fileUrl` option is a good way to do that. But I don't know how to use `python manage.py` or `higlass-manage` to ingest data with `fileUrl`. Has it already been accomplished?

pkerpedjiev:
Unfortunately, you can't use `fileUrl` through `higlass-manage`. It was meant as a way for developers to create web applications that access data that is not served by a higlass-server.

To be honest, if you're just looking for one dataset, the best way to get it would be to download it and ingest it locally.

pkerpedjiev:
But that's actually not a bad idea. Maybe we can add a way to ingest data that is accessible on a different higlass server 🤔 I'll have to think about that a little bit.

PR #430, opened 2018-11-25
Expose version

flekschas:
Can be useful for debugging

## Description

Adds `version` to HgApi.

> Why is it necessary?

To be able to determine the version of Hglib dynamically with JS. Useful for the version dialog in HGA.

## Checklist

- ~~[ ] Example added or updated~~
- ~~[ ] Screenshot for visual changes (e.g. new tracks)~~

### DIFF
--- a/app/scripts/api.js
+++ b/app/scripts/api.js
@@ -39,6 +39,9 @@ const createApi = function api(context, pubSub) {
        * @return {string} Version number
        */
       get version() {
+        // Note, `VERSION` is exposed by webpack across the entire app. I.e.,
+        // it's globally available within the build but not outside. See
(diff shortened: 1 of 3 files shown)


app/scripts/api.js
```diff
@@ -34,6 +34,14 @@ const createApi = function api(context, pubSub) {
     publish: apiPubSub.publish,
     // Public API
     public: {
+      /**
+       * HiGlass version
+       * @return {string} Version number
+       */
+      get version() {
+        return VERSION;
```
pkerpedjiev: Can we get a comment explaining where VERSION is originally defined?

flekschas: Good point! Will add.

PR #439, opened 2018-11-26
"cross-rule" is a another special case which is not waiting for anything

mccalluc:
## Description

- Add tests for trivial viewconfs. (Though maybe these are too trivial... should there be an error in cases like these?)
- Add another special case to test-helpers.js: `cross-rule` is another instance where we do not need to wait for tiles. (Would there be any way to determine from the outside if a particular track type is one we should wait for?)

Why is it necessary?
- I stalled out on the cross-rule svg export earlier because the test I wanted to write kept failing, and I didn't understand why.

## Checklist

- [X] Unit tests added or updated
- [ ] Updated CHANGELOG.md

### DIFF
--- a/app/scripts/utils/test-helpers.js
+++ b/app/scripts/utils/test-helpers.js
@@ -137,6 +137,7 @@ export const isWaitingOnTiles = (hgc) => {
       || track.track.type === 'horizontal-1d-annotations'
       || track.track.type === 'vertical-1d-annotations'
       || track.track.type === '2d-chromosome-annotations'
+      || track.track.type === 'cross-rule'
     ) continue;
 
     if (trackObj.originalTrack) { trackObj = trackObj.originalTrack; }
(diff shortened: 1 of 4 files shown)


pkerpedjiev:
> Would there be any way to determine from the outside if a particular track type is one we should wait for?
Maybe if the track doesn't have `server` and `tilesetUid` set then it doesn't have to wait for tiles?

pkerpedjiev:
Things are definitely in a weird state. Take `simpleCenterViewConfig` for example. It contains this track definition:

```
          {
            type: 'top-axis',
            local: true,
            orientation: '1d-horizontal',
            name: 'Top Axis',
            thumbnail: {},
            defaultOptions: {},
            server: '',
            tilesetUid: 'UKZhWC_FSCG1xHlN0c2bqQ',
            uid: 'd8focpO5TKKd2TB8QONd9w',
            options: {},
            width: 20,
            height: 20,
            position: 'top'
          }
```

A `top-axis` should have neither a server nor a tilesetUid.  `thumbnail` should not be part of the viewconf and neither should `defaultOptions`. 

Maybe see if the empty server field is causing this error? But those fields should definitely not be exported in the viewconf. There's code to remove invalid fields from a JSON here: https://github.com/higlass/higlass/blob/57700e542c5730f4e4218ff2200ad67958144892/app/scripts/HiGlassComponent.js#L2429

That stuff should probably never end up in the internal viewconf in the first place but that's a topic for another discussion.

mccalluc:
That fixed it. I'll go ahead and merge.

PR #440, opened 2018-11-26
Clear, easy to navigate examples

mccalluc:
## Description

- Examples which are viewconf-only have been reduced to their json, eliminating copy and paste html in the examples.
- This allows `svg.html` to load them all easily, and demonstrates several unreported bugs.
- Rather than copying and pasting filenames, `examples.html` now links to all the examples: Although not automatically generated, one-liners in the comments should make it easy to keep up to date.
- api examples have been moved to a subdirectory to keep the top-level simple.
- and now the higlass.io examples are also available here, too.

Why is it necessary?

- Generating SVGs from the other example viewconfs was the primary motivation.

To the best of my knowledge, nothing outside `examples` depends on the contents of this directory. If that's not true, I may need to reconsider this.

(The one-liners are questionable... Client-side code could probably do the job... but I think having a list of the examples actually checked in is not a bad thing. Or I could make one script to generate the whole thing, instead of the one-liners?)

screen shot 2018-11-26 at 6 30 38 pm

## Checklist

- [X] Example added or updated
- [ ] Updated CHANGELOG.md


pkerpedjiev:
This looks fantastic and extremely helpful!

If we want to link to these examples from the docs, we'll have to copy the latest hglib.js to the S3 bucket hosting the docs, right? Would it make sense to include that in this PR?

As far as I can tell, the hosted demos links (e.g.  like in this section https://github.com/higlass/higlass/blob/develop/docs/track_types.rst#gene-annotations) are broken in the online docs.

pkerpedjiev:
The rst on GitHub doesn't matter. The rendered documentation does. And that should be fine with relative paths as long as hglib is copied over to the root location.

What is the S3 url again? I'm becoming more and more convinced that we need to switch over to that ASAP.

PR #443, opened 2018-11-27
Provide fake pubSub by default for convenience

flekschas:
## Description

By default a fake pubSub object is provided so that ChromosomeInfo is backwards compatible

Why is it necessary?

For convenience


mccalluc:
(actually, I already merged my pr based on this, so this is already in.)

PR #448, opened 2018-11-28
Pkerpedjiev/fix build

pkerpedjiev:
## Description

This fixes the build issue that miniconda was causing by forcing it to install to the specified directory even if it already exists. Miniconda is used to provide python3 for building the docs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ script:
     - ./travis_build_test.sh
 before_deploy:
     - wget http://repo.continuum.io/miniconda/Miniconda3-4.3.21-Linux-x86_64.sh -O miniconda.sh
-    - bash miniconda.sh -b -p -f $HOME/miniconda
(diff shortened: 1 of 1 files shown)


pkerpedjiev:
> Do you consider a successful doc build as part of a successful PR... or is it only necessary when we're actually going to deploy new docs to s3? If it is part of every PR, then we might want to have it under the travis_build_test.sh umbrella, or if not, we might might not need to run this part every time.

Yes. Each PR should successfully build everything. If we don't enforce docs building, then when it comes time for a release, we'll have to go back and determine where the problem arose.

> If we need both python and node, it's possible that we might want to change the top level language in .travis.yaml, unless conda is really necessary to get the doc build to work?

I briefly looked into this but couldn't find a good way to specify both python and npm. If you can find a better way to do it, feel free to do so but unless there's an error, it doesn't seem like it's particularly urgent to do that.

PR #454, opened 2018-12-04
related tools diagram

mccalluc:
## Description

*Wait until the diagram is actually what you want!*

- Show the current version of the diagram, and link to google draw so we'll know where it is and can edit.
- The downside of not checking in an exported image is that we won't keep the history, but seems much more maintainable.

Why is it necessary?
- We have a lot of moving parts, and it's hard to know how they're related.

## Checklist

- [X] Documentation added or updated
- [ ] Updated CHANGELOG.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,8 @@ const api = hglib.createHgComponent(
 ```
 ### Related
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -83,6 +83,8 @@ const api = hglib.createHgComponent(
 ```
 ### Related
 
+[![diagram of related tools](https://docs.google.com/drawings/d/e/2PACX-1vSCiCzfQ8FEyHPFSq7jJD6XmzC760xH1Zr4FIcCMzFmqAlrmYEBMId8gM42uz0okmvuEaxetyPPZ9VG/pub?w=960&h=720)](https://docs.google.com/drawings/d/1Xedi5ZRtbRdt2g20qpl_lWs4BMqc2DKZ2ZOoJvpHw9U/edit)
```
pkerpedjiev: This image doesn't have nearly the resolution necessary to make out what's in it.

mccalluc: @pkerpedjiev : I've deleted all but the middle view (so 960x720 is sufficient, or maybe even a bit much). Also added a key for the colors. Merge if it looks good to you?

PR #457, opened 2018-12-05
Mccalluc/minimal linting

mccalluc:
## Description

- Ran `--fix` on the entire codebase
- Reverted chunks in the react code that had been corrupted by `--fix`
- Make a new `.eslintrc-minimal` and `.eslintignore-minimal`, and in `.eslintrc-minimal` ignore every error we still have
- Add a second call to eslint in `travis_build_test.sh`: This call will just do the minimum linting across the entire codebase. 

Why is it necessary?
- This is a horizontal approach to linting, and complements the vertical approach taken earlier.
- I do not want this to be a permanent addition, but it's a way of getting some minimal rules enforced across the board now, before we have the chance to fix each individual file.
- 63 rules are being ignored right now. Many of them might be trivial to fix, and I'd like to do that, but this PR is huge as it is, and if I'm making changes that require even a little judgement, I want that them to be in their own PRs so they can be reviewed more carefully.

## Checklist

- [X] Unit tests added or updated
- [ ] Updated CHANGELOG.md


mccalluc:
(Going to fix the indentation on more of the react code... new commit soon.)

mccalluc:
I've [branched from this](https://github.com/higlass/higlass/compare/mccalluc/minimal-linting...mccalluc/and-fix-some-simple-linting-errors?expand=1) and started making a few simple fixes that are now enforced globally. I'll do a little bit more work there today and then leave it: I mostly wanted to give you an example of how small incremental fixes could be made going forward.

flekschas:
Any thoughts on setting up git hooks in parallel so we avoid accidentally breaking the TravsiCI build etc? Just an idea. I can help setting up the hook once the minimal linting runs through (setting it prior that does not make sense). Of course this could also be another PR

pkerpedjiev:
@mccalluc This is great! As the biggest linting villain around, it'll help keep my wayward ways in check and that's fantastic.

@flekschas Are you suggesting that we run the linter after every git commit? How long does a typical linting run take? If it's more than a second it's going to be annoying.

app/scripts/d3-context-menu.js
```diff
@@ -13,8 +13,10 @@ function contextMenu(menu, opts) {
   let initialPos = null;
   let parentStart = null;
 
-  let openCallback,
-    closeCallback;
+  let openCallback;
+
+
```
pkerpedjiev: Why two empty lines here?

mccalluc:
One build failed: filed #464. Restarting Travis.

app/scripts/d3-context-menu.js
mccalluc: No idea: It's just what `--fix` did. I think I'll grep for sequences of more than 1 newline in a row, and fix them all, but after this.

mccalluc:
One build failed: filed #464. Restarting Travis.

mccalluc:
@flekschas: Besides the speed issue, right now the horizontal linting is a *very* low bar. I think it might be good during PRs to ask if the files touched have been added to vertical linting coverage, but I don't think that's something we can enforce right now: Getting a particular file to lint at the higher standard might be a hassle you don't need right then.

flekschas:
@pkerpedjiev @mccalluc Yes I was suggesting to run a linter on every staged file that's suppose to be committed. The effort is minimal and it's generally very fast (few ms) unless you're committing files every few seconds or more than 500 at once speed is really not an issue here. The reason I broad this up is that we probably want to spot having these lint PRs in the future and while 1 time linting is nice, it doesn't stop us from commited files that will subsequently fail on TravisCI (happened to me after merged a couple of times now)

pkerpedjiev-zymergen:
As long as it's fast, I don't mind. I presume we all have linters set up in our editors so it shouldn't even get to the hooks.

Issue #459, opened 2018-12-05
"Checks the axis texts" failed once on Travis


mccalluc:
Just making a record of this in case we hit it again and wonder if there is any history.

https://travis-ci.org/higlass/higlass/builds/463992098#L3386
```
05 12 2018 19:09:27.302:ERROR [Chromium 62.0.3202 (Ubuntu 0.0.0) | Simple HiGlassComponent, Axis texts | Checks the axis texts]: Script error. thrown
05 12 2018 19:09:27.302:ERROR [Chromium 62.0.3202 (Ubuntu 0.0.0) | Simple HiGlassComponent, Axis texts | Checks the axis texts]: TypeError: Cannot read property 'getTrackObject' of null
    at getTrackObjectFromHGC (webpack://%5Bname%5D/./scripts/utils/get-higlass-components.js?:25:51)
    at UserContext.eval (webpack://%5Bname%5D/../test/AxisTests.js?:44:53)
```

PR #460, opened 2018-12-05
Mccalluc/and fix some simple linting errors

mccalluc:
## Description

- Fixing some of the simplest linting errors across the board

There was one travis error which was not repeated. I've filed #459 just to have a record of it.

### DIFF
--- a/app/scripts/DataFetcher.js
+++ b/app/scripts/DataFetcher.js
@@ -336,9 +336,6 @@ export default class DataFetcher {
           }
         }
 
-        const min_value = Math.min.apply(null, dataSlice);
-        const max_value = Math.max.apply(null, dataSlice);
-
         const newTile = {
           min_value: Math.min.apply(null, dataSlice),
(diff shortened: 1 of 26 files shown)


app/scripts/DataFetcher.js
```diff
@@ -219,7 +219,7 @@ export default class DataFetcher {
    * Extract a slice from a matrix at a given position.
    *
    * @param {array} inputData: An array containing a matrix stored row-wise
-   * @param {array} arrayShape: The shape of the array, should be a two element array e.g. [256,256].
+   * @param {array} arrayShape: Shape of the array; should be a two element array e.g. [256,256].
```
pkerpedjiev: I never did know where to use semicolons when writing in English.

flekschas: I always thought of it as being a longer pause / stronger separation than a comma but less then a dot. 🤔 (Sorry this is a random comment)

PR #461, opened 2018-12-05
Flekschas/js api new cursor location

flekschas:
## Description

There's a new event one can listen to called `cursorLocation` which returns the location of the cursor when hovering a higlass instance.

Why is it necessary?

Because `mouseMoveZoom` only works when a track is able to extract data under the mouse cursor. Right now only the HeatmapTiledPixiTrack supports that. but in order to just get the location of the mouse cursor a track doesn't have to do anything. Ideally `mouseMoveZoom` is renamed to something like `trackDataUnderCursor` but that's another PR.

When the event is triggered it's publishing a subset of `HiGlassComponent.mouseMoveHandler()`. E.g.,

```javascript
{
  absX: 727
  absY: 545
  dataX: 1434506618.3999963
  dataY: 1193639169.4500015
  isFrom2dTrack: true
  isFromVerticalTrack: false
  relTrackX: 702
  relTrackY: 33
  relX: 717
  relY: 535
}
```

And this works when mousing over *any* track.

## Checklist

- [ ] Updated CHANGELOG.md


pkerpedjiev:
```
  isFrom2dTrack: true
  isFromVerticalTrack: false
```

Why not have something like:

```
  sourceTrackOrientation: [2d | vertical | horizontal],
  tracks: [track1Obj, track2Obj]
```
?

Where `tracks` would contain the track objects that are below the cursor? 

Also, is there a reason you don't include the tracks under the cursor already?

PR #462, opened 2018-12-05
Flekschas/add margin to axis

flekschas:
## Description

One can add some margin to axes via `track.options.axisMargin`

Why is it necessary?

Sometimes other UI elements hide the very left, right, top, or bottom position of the axis. In that case it's handy to be able to push the label inwards or outwards.

screen shot 2018-12-05 at 4 58 57 pm

On a side note, this PR fixes two minor issues with right aligned axis (called LeftAxis)

## Checklist

- [x] SVG rendering works as well
- [x] Updated CHANGELOG.md


docs/examples/apis/axis-margin.html
```diff
@@ -0,0 +1,354 @@
...
+        static: false
+      },
+    }
+  ],
+};
+const api = hglib.viewer(
+  document.getElementById('demo'), viewConfig, { bounded: true },
+);
+</script>
+</html>
```
mccalluc: Does this do anything special with the API? If not, I'd prefer to only have the json file checked in, but I could be missing something.

flekschas: I will remove this prior to merging. My only beef with the JSON file right now is that it's more work / less convenient for quick testing. Therefore, I added an intermediate html file until the PR is fully ready.

flekschas:
@mccalluc You are right. I did not care about SVG rendering. Will add a checkmark and implement it at some point in the future.

flekschas:
@mccalluc You are right. I did not care about SVG rendering. Will add a checkmark and implement it at some point in the future.

pkerpedjiev:
Good catch @mccalluc. If I had realized that I wouldn't have given it a stamp of approval.

Much easier to add the SVG export code while this is still fresh in your mind Fritz. Otherwise it's another issue we have to track and make sure to fix before (or after) somebody complains.

flekschas:
It would be great if we could support a 2-step process, i.e., to have a way to indicate SVG exportability. I know it would be the best to have everything in place but time's a limiting factor. Hence, it would be great to be able to ship features even if not all use cases are supported.

Nobody can complain if it's advertised as WebGL rendering only :)

Having said that, I might get around implementing SVG support at some point but I am not sure when.

PR #476, opened 2018-12-11
Properly clear mouse position indicator

flekschas:
## Description

Properly clear mouse position indicator on mouseLeave, on blur, and when not hovering any track

Why is it necessary?

Before, the mouse position wasn't cleared which is annoying when there are multiple HiGlass instances as one always saw the mouse pos when leaving the root dom element which is uninformative.

## Checklist

- [x] Updated CHANGELOG.md


app/scripts/utils/show-mouse-position.js
```diff
@@ -41,6 +41,8 @@ const showMousePosition = (
   // Graphics for cursor position
   const graphics = new PIXI.Graphics();
 
+  const clearMousePosition = () => { graphics.clear(); };
```
pkerpedjiev: Why is this called `clearMousePosition`? It clears the graphics, not the mouse position.

app/scripts/utils/show-mouse-position.js
```diff
@@ -50,7 +52,7 @@ const showMousePosition = (
    * @param  {Boolean}   isNoClear  If `true` do not clear the graphics.
    */
   const drawMousePosition = (mousePos, isHorizontal, isNoClear) => {
-    if (!isNoClear) graphics.clear();
+    if (!isNoClear) clearMousePosition();
```
pkerpedjiev: To follow up on the comment above, why is it necessary to put the graphics.clear() call in a separate function. IMO it's a little misleading.

If you want to explain that the `graphics.clear()` call is used to clear the mouse position, I think a comment would be better suited.

flekschas:
```suggestion
    if (!isNoClear) graphics.clear();
```

Leftover. Good catch.

app/scripts/utils/show-mouse-position.js
flekschas:
```suggestion
```

Leftover. Good catch.

flekschas:
@pkerpedjiev Changed the name to `clearGraphics` :) Shall I merge it now if the TravisCI builds succeed or shall we wait first for v1.3 to come out?

pkerpedjiev:
You can merge it now. It's a bug fix and not very risky.

PR #482, opened 2018-12-18
enforce consistent-return

mccalluc:
Though `null` might be better than `undefined`? It would make it clear that this is a value we chose to return, rather than just falling out the bottom of a function.

## Description

- add `return undefined` as needed

## Checklist

- [ ] Updated CHANGELOG.md


mccalluc:
The push build failed on "APIs are independent" ... I'll restart that test.

PR #485, opened 2018-12-18
enforce no-prototype-builtins

mccalluc:
but if we would never shadow Object.prototype methods, this is excessive, and it would be better just to ignore the rule.

Thoughts? I'd be happy just to disable the rule, too.

### DIFF
--- a/app/scripts/HiGlassComponent.js
+++ b/app/scripts/HiGlassComponent.js
@@ -881,7 +881,7 @@ class HiGlassComponent extends React.Component {
    *    Event handler is called with parameters (xScale, yScale)
    */
   addDraggingChangedListener(viewUid, listenerUid, eventHandler) {
-    if (!this.draggingChangedListeners.hasOwnProperty(viewUid)) {
+    if (!Object.prototype.hasOwnProperty.call(this.draggingChangedListeners, viewUid)) {
       this.draggingChangedListeners[viewUid] = {};
     }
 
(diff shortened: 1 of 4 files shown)


app/scripts/HiGlassComponent.js
```diff
@@ -881,7 +881,7 @@ class HiGlassComponent extends React.Component {
    *    Event handler is called with parameters (xScale, yScale)
    */
   addDraggingChangedListener(viewUid, listenerUid, eventHandler) {
-    if (!this.draggingChangedListeners.hasOwnProperty(viewUid)) {
+    if (!Object.prototype.hasOwnProperty.call(this.draggingChangedListeners, viewUid)) {
```
flekschas: Just out of curiosity, is there a benefit of this over a simple `this.draggingChangedListeners[viewUid]`? I've never seen `Object.prototype.hasOwnProperty.call` before (doesn't mean anything but it looks much more complicated than it needs to be in my eyes).

flekschas:
In my opinion I would solve all the instances with `hasOwnProperty` with a simple brackets lookup. E.g., instead of `x.hasOwnProperty(test)` I would do `x[test]`. Otherwise I find https://eslint.org/docs/rules/no-prototype-builtins to be too excessive but I don't mind if we apply it either.

mccalluc:
I would worry about using `x[test]` to check for key existence... What if it's there, and the value is false? Shouldn't be the case for callbacks, though. I'm happy to just disable that rule. New PR in soon, and I'll close this one.

PR #493, opened 2018-12-20
enforce import/no-extraneous-dependencies

mccalluc:
## Description

- Move `test-helpers` from `app/scripts/utils` to `test/utils`.
- Update imports accordingly.
- Enforce `import/no-extraneous-dependencies`.

## Checklist

- [ ] Updated CHANGELOG.md


mccalluc:
@pkerpedjiev : The simplest thing would be to disable eslint on this line so that it's still available. (The right thing, which probably isn't worth the trouble, is to split out these utils to their own package, so it would be clear that tests and plugin tracks need it, but nothing else relies on it.)

pkerpedjiev:
You mean the line where they're exported in `hglib.js`? That'd be fine with me.

mccalluc:
Replaced by #502.

PR #507, opened 2019-01-07
Mccalluc/no unused vars

mccalluc:
## Description

- Fix `no-unused-vars`
- After fixing a few, decided a better approach is to keep the destructuing, and ignore individual lines, but this could be argued.

## Checklist

- [ ] Updated CHANGELOG.md


app/scripts/GenomePositionSearchBox.js
```diff
@@ -502,7 +502,7 @@ class GenomePositionSearchBox extends React.Component {
     for (let i = 0; i < valueParts.length; i++) {
       if (valueParts[i].length === 0) { continue; }
 
-      const [chr, pos, retPos] = this.searchField.parsePosition(valueParts[i]);
+      const retPos = this.searchField.parsePosition(valueParts[i])[2];
```
pkerpedjiev: Isn't this contrary to the "destructure with disabled warning philosophy"?

mccalluc: A little: I was more thinking of the cases where we might also be getting `pos` but not `chr`, where it would have to be split over multiple lines if we're not using destructuring... ie, it's not that access by index is bad, but changing one tidy line with an `eslint-disable` into two or more separate assignments feels awkward.

Eh. It depends on context. If you have a strong preference, let me know, but I'm comfortable with this.

PR #512, opened 2019-01-08
Enforce no-array-index-key

mccalluc:
... but ignore where there is not a better option

## Description

- Enforce no-array-index-key.
- Add one block-disable. (Line disables are tricky in JSX.)

If you'd prefer to disable the rule globally (perhaps the problem is obvious enough so you probably know what you're doing if you do violate it?) I could do that instead. Your call.

## Checklist

- [ ] Updated CHANGELOG.md


mccalluc:
On the PR build: `* APIs are independent : failed`

PR #521, opened 2019-01-09
Elevate warnings to errors

mccalluc:
## Description

- These are all the potential warnings which we are not actually triggering. If the warning mechanism is to be meaningful, the list will need to stay short. Some of these might be things we actually care about, and would like to be sure we catch violations. Others of them may not be... when we encounter a violation in the future, we could decide then to turn off the rule.

This isn't critical, but thought it wouldn't hurt to propose it.

## Checklist

- [ ] Updated CHANGELOG.md


pkerpedjiev:
So changing these to errors is the equivalent to "this suggestion is now becoming law"?

mccalluc:
> "we're following these suggestions so let's just make them hard and fast rules (until such a time as we need to break them)"?

Yeah: When a violation does crop up, our response could be
- "Good catch! Glad that rule was on."
- "That's silly! I'll disable that rule as part of my PR."
- "Good point -- but what I'm doing now is special, so I'll disable for just this line."

PR #524, opened 2019-01-12
Flekschas/overlay tracks 1d and 2d

flekschas:
## Description

This PR adds support for 1D vertical and 2D overlay tracks. Overlay tracks also support the option `fillOpacity` now.

> Why is it necessary?

Because previously overlay tracks were only visible for 1D horizontal tracks.

screen shot 2019-01-11 at 5 04 36 pm

## Checklist

- [ ] ~Updated CHANGELOG.md~

### DIFF
--- a/app/scripts/OverlayTrack.js
+++ b/app/scripts/OverlayTrack.js
@@ -25,7 +25,7 @@ class OverlayTrack extends PixiTrack {
       const orientation = this.options.orientationsAndPositions[i].orientation;
       const position = this.options.orientationsAndPositions[i].position;
 
-      if (orientation === '1d-horizontal') {
+      if (orientation === '1d-horizontal' || orientation === '2d') {
         const xPos = this.position[0]
           + position.left
           + this._xScale(this.options.extent[0][0]);
@@ -39,7 +39,7 @@ class OverlayTrack extends PixiTrack {
         graphics.drawRect(xPos, yPos, width, height);
       }
 
-      if (orientation === '1d-vertical') {
+      if (orientation === '1d-vertical' || orientation === '2d') {
         const xPos = this.position[0] + position.left;
         const yPos = this.position[1]
           + position.top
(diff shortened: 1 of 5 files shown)


flekschas:
Do you have any ideas on how to test a PIXI draw call? The original overlay tracks seem to have never been tested.

pkerpedjiev:
I know, I'm really sorry for pawning the testing off on your PR. I think when I first implemented it there were no tests. 

You could use `PIXI.Rectangle` and attach it to the scene: http://pixijs.download/dev/docs/PIXI.Rectangle.html

Another option would be to use a spy function: https://jasmine.github.io/2.0/introduction.html#section-Spies

Dynamically replace the track's draw function with a spy and make sure it gets called? 

Even something as simple as making sure the track object is created and it's properly positioned would be helpful.

flekschas:
No worries, having tests totally makes sense. I was just wondering what the best way would be to test it.

flekschas:
@pkerpedjiev I've added a basic test that checks if the track was defined, created, and initialized. Would that be sufficient for now? There's probably a way to figure out if and where PIXI drew some overlays on the canvas but I am not sure how to implement that. The current solution should at least ensure that the track exists if it's defined in the viewconf.

flekschas:
And here we go again... The PR runs successfully but the push fails because of some timeout.

@pkerpedjiev Any idea what the issue here might be:

```
...
17 01 2019 21:29:04.722:DEBUG [middleware:source-files]: Requesting /base/test/view-configs/viewport-projection.js?b17870177e392a07ff0160bae875555c947b66d4 /
17 01 2019 21:29:04.722:DEBUG [middleware:source-files]: Fetching /home/travis/build/higlass/higlass/test/view-configs/viewport-projection.js
17 01 2019 21:29:04.723:DEBUG [web-server]: serving (cached): /home/travis/build/higlass/higlass/test/view-configs/viewport-projection.js
17 01 2019 21:29:08.154:WARN [Chromium 62.0.3202 (Ubuntu 0.0.0)]: Disconnected (1 times), because no message in 10000 ms.
17 01 2019 21:29:08.155:ERROR [Chromium 62.0.3202 (Ubuntu 0.0.0)]: Disconnected, because no message in 10000 ms.
```

Assuming this is again some kind of issue with a race condition or a hiccup my best guess is to just restart the test?

PR #525, opened 2019-01-14
Flekschas/overlay chrom grid

flekschas:
## Description

Add support for different types of overlay track. Specifically, this PR adds `type: chromosome-grid`.

screen shot 2019-01-14 at 1 34 57 pm

It even works with pan & zoom 🤓
jan-14-2019 14-07-01

> Why is it necessary?

It's not! But it's much more convenient to just have to specify one overlay track than having to combine a chrom grid track with every other track.

> Wait a moment! Before we cool, show me how you define those overlay grids buddy!

Of course!

```
      overlays: [
        {
          uid: 'overlay-chroms',
          type: 'chromosome-grid',
          includes: ['chroms-top', 'genes-top', 'chroms-left', 'genes-left', 'center'],
          chromInfoPath: '//s3.amazonaws.com/pkerp/data/hg19/chromSizes.tsv',
          options: {
            lineStrokeWidth: 1,
            lineStrokeColor: 'grey'
          }
        }
      ]
```

The track definition is almost identical to the normal chromosome-2d-grid track except that it needs to include the `include` property that we know and love from overlay tracks.

## Checklist

- [x] Updated CHANGELOG.md
- [x] SVG rendering works


flekschas:
Yes, good call. I forgot about the tests :)

What do you mean with "will it be exported as an overlay or as a bunch of individual tracks?". Do you refer to SVG exporting? Classic overlay tracks so far never implemented any SVG export but I would image they are exported just like a normal track and not multiple tracks.

flekschas:
Affirmative sir! Will do! 🤓

flekschas:
@pkerpedjiev Tests and docs are added

flekschas:
SVG rendering works now:

screen shot 2019-01-24 at 1 02 20 pm

PR #530, opened 2019-01-17
Expand the documentation of multivec files

pkerpedjiev:
## Description

More documentation about multivec files

## Checklist

- [o] Unit tests added or updated
- [o] Updated CHANGELOG.md

### DIFF
--- a/docs/data_preparation.rst
+++ b/docs/data_preparation.rst
@@ -640,11 +640,12 @@ Other Data (multivec)
 ---------------------
 
 Multivec files are datatype agnostic. For use with generic data, create a
-`segments` file containing the maximum value for each segment. A segment is an
+`segments` file containing the length of each segment. A segment is an
 arbitrary set of discontinuous blocks that the data is partitioned into. In the
 case of genomics data, segments correspond to chromosomes. If the
-data has no natural grouping, one segment can be used which contains the
-maximum x value in the dataset:
+data has no natural grouping, it can all be lumped into one "segment"
+which is wide enough to accommodate all the data points. Below is an
+example of a dataset grouped into two "segments".
 
 .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


docs/data_preparation.rst
```diff
@@ -641,20 +641,31 @@ Other Data (multivec)
...
 .. code-block:: bash
 
-    segment_name    start   end value
-    segment1    10  100 5
+    |---------------|------------------------------|
+         segment1               segment2
+
+The individual datapoints should then be formatted as in the block below. Each
+row in this file corresponds to a column in the displayed matrix. Each ``value``
+is a row in the matrix.
```
flekschas: What is this matrix you are referring to? Are you saying that 

```
segment1           10  100       5      15       20
```

is translated into a stacked bar that covers the range [10-100] and has 3 segments defined by the 3 value columns?

I am still confused how this row defining `segment1` with the list above where it says `segment1 20000 `

docs/data_preparation.rst
```diff
@@ -641,20 +641,31 @@ Other Data (multivec)
 
 Multivec files are datatype agnostic. For use with generic data, create a
 `segments` file containing the maximum value for each segment. A segment is an
-arbitrary set of discontinuous blocks that the data is partitioned into. If the
+arbitrary set of discontinuous blocks that the data is partitioned into. In the
+case of genomics data, segments correspond to chromosomes. If the
 data has no natural grouping, one segment can be used which contains the
 maximum x value in the dataset:
```
flekschas: > one segment can be used which contains the maximum x value in the dataset

I am not sure I get what this means. What does the 20000 and 40000 in the list below refer to? Maybe I am just missing context but I am not sure what the segment has to do with some maximum value.

pkerpedjiev: I updated the docs. This was originally designed with chromosomes in mind. So the 20000 and 40000 are the chromsizes. They're the lengths that segments will occupy along the x-axis. So if you have two segments:

segment1 20000
segment2 40000

And a datapoint at `segment 2 0`, it will actually get drawn at position 20000 on the x-axis because the first 20000 is occupied by the first segment. Does that make sense?

flekschas: Oh so `segment1: 10 -100 [5, 15, 20]` would be a data point in `segment1` that spans `[10-100]` and has 3 values attached to it? Just to be sure I get the idea: would this datapoint be ultimately be merged with other data points of segment1 or would it be visualized irrespectively of the segment it's lying in? I guess my question is what exactly is the segment definition good for in terms of the visual representation?

PR #536, opened 2019-01-25
Add the ability to use the glyph color for the stroke color

pkerpedjiev:
## Description

Give the line and bar tracks the ability to use the color of the stroke or fill from the plot for their coloring. If the value of `labelColor` is `[glyph-color]`, then take the value `lineStrokeColor` or `barFillColor`.

image

Why is it necessary?

This is helpful when there's multiple lines and we want to be able to match labels to lines.

## Checklist

  - Updated track_types.rst to add information about the newly available option
- [x] Updated CHANGELOG.md


flekschas:
Out of curiosity: what is the new option? And is there any difference to having the same color for `labelColor`, `lineStrokeColor`, and `barFillColor`?

PR #538, opened 2019-01-25
Fixes overlay tracks to allow for multiple extent values and 4-tuples.

mildewey:
## Description

A) Overlay tracks were augmented to allow for extent to draw multiple ranges. This was an intended feature that was not implemented properly. B) It was also augmented to allow for drawing x and y ranges differently for the "same" extent. This may or may not have been intended initially, but allows for flexibility in how overlay tracks are drawn.

Why is it necessary?
The alternative was to make a large number of overlay tracks and configure each identically. This seemed obnoxious and counterintuitive.

Fixes #___

## Checklist

- [x] Updated CHANGELOG.md
higlassoverlay


pkerpedjiev:
Could you please fill out the PR request template? The changes look fine and they seem to work when I tried them.

app/scripts/OverlayTrack.js
```diff
@@ -21,38 +49,19 @@ class OverlayTrack extends PixiTrack {
...
-          + this._yScale(this.options.extent[0][0]);
-        const width = position.width;
-        const height = this._yScale(this.options.extent[0][1])
-          - yPos
-          + position.top
-          + this.position[1];
-
-        graphics.drawRect(xPos, yPos, width, height);
+      if (op.orientation === '1d-vertical' || op.orientation === '2d') {
+        this.options.extent.forEach(extent => this.drawVerticalOverlay(
```
flekschas: Will `this.options.extent.forEach` break if `extent` is configured incorrectly? A fix could be to test at the beginning whether extent is an array, i.e., `this.extent = this.options.extent || []`

flekschas: (Same in the above loop)

app/scripts/OverlayTrack.js
```diff
@@ -11,6 +11,34 @@ class OverlayTrack extends PixiTrack {
     this.drawnRects = {};
   }
 
+  drawHorizontalOverlay(graphics, position, extent) {
```
flekschas:
```suggestion
  drawHorizontalOverlay(graphics, position, extent = [0, 1]) {
```

app/scripts/OverlayTrack.js
```diff
@@ -11,6 +11,34 @@ class OverlayTrack extends PixiTrack {
...
+    const height = position.height;
+    const width = this._xScale(extent[1])
+      - xPos
+      + position.left
+      + this.position[0];
+
+    graphics.drawRect(xPos, yPos, width, height);
+  }
+
+  drawVerticalOverlay(graphics, position, extent) {
```
flekschas:
```suggestion
  drawVerticalOverlay(graphics, position, extent = [0, 1]) {
```

flekschas:
Looks fine. I've just left a few minor comments regarding robustness. Also, please add or extend:

- the existing example
- CHANGELOG.md

app/scripts/OverlayTrack.js
mildewey: Resolved by checking if extent exists and is long enough and doing nothing if either is bad.

app/scripts/OverlayTrack.js
mildewey: Fixed by making sure we have extent as an array.

PR #552, opened 2019-02-01
Add an example of a tile response

pkerpedjiev:
## Description

Update the docs to include a sample tile response.

Why is it necessary?

So one doesn't have to open the developer console to see what tile responses look like.

## Checklist

- [ ] Updated CHANGELOG.md

### DIFF
--- a/docs/higlass_server.rst
+++ b/docs/higlass_server.rst
@@ -174,6 +174,21 @@ To modify a tileset name, specify the tileset `uuid` in the URL, use the `PATCH`
 
     curl --user ${username}:${password} --request PATCH --header "Content-Type: application/json" --data '{"name":"new_name_of_tileset"}' http://localhost:8000/api/v1/tilesets/${uuid}/ 
       
+Tile JSON Response Format
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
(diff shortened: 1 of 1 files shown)


docs/higlass_server.rst
```diff
@@ -174,6 +174,21 @@ To modify a tileset name, specify the tileset `uuid` in the URL, use the `PATCH`
...
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Tiles returned by the server vary according to the data type but
+all are indexed by their tile id:
+
+.. code-block:: bash
+
+  {
+    "OHJakQICQD6gTD7skx4EWA.3.2": [
+      { "uid": "US2sjy_8SlGuy-0iSshcDQ", "importance": 457.0, "fields": [...] }
```
flekschas: Are `uid`, `importance`, and `fields` mandatory properties? If so could you add a note? I think this would be useful to know.

pkerpedjiev: Yes they are, uid is necessary because if an annotation spans multiple tiles, it will be returned with all of them. Importance prioritizes them and fields contains the data. I'll that to the document.

flekschas: Awesome thanks!

PR #563, opened 2019-02-15
Flekschas/global mouse position

flekschas:
## Description

This pull request allows broadcasting the mouse position across browser tabs.

> Why is it necessary?

It's not necessary but it's a cool feature when working with HiGlass in multiple browser tabs simultaneously, e.g., when using multiple computer screens, or when working with multiple HiGlass instances.

See http://localhost:8080/others/global-mouse-position.html for a demo. The global mouse position event support 2 use cases:

1. It allows broadcasting events between different HiGlass instances (in the demo the two HG views related to two different HG instaces)
2. Between browser tabs. To see how this works open up http://localhost:8080/others/global-mouse-position.html in a second browser tab.

feb-22-2019 17-53-02

## Checklist

- [x] Updated CHANGELOG.md


app/scripts/api.js
```diff
@@ -45,6 +45,44 @@ const createApi = function api(context, pubSub) {
...
+       *   `options.showMousePosition = true`.
+       */
+      setShowGlobalMousePosition(isShowGlobalMousePosition = false) {
+        self.setShowGlobalMousePosition(
+          isShowGlobalMousePosition
+        );
+      },
+
+      /**
+       * Convinience function to enable / disable the global mouse position
```
pkerpedjiev:
```suggestion
       * Convenience function to enable / disable the global mouse position
```
Spelling

app/scripts/utils/show-mouse-position.js
```diff
@@ -75,7 +79,7 @@ const showMousePosition = (
    * @param  {Object}  e  Event object.
    */
   const mouseMoveHandler = (event) => {
-    if (!event.hoveredTracks.length) {
+    if (event.noHoveredTracks) {
```
pkerpedjiev: To me the alternative (!event.hoveredTracks.length) is more versatile because we don't have to keep track of both `event.hoveredTracks` and `event.noHoveredTrack`. I think the text "no hovered tracks" is helpful but as a comment rather than as a variable.

app/scripts/TiledPlot.js
```diff
@@ -1987,7 +1987,8 @@ class TiledPlot extends React.Component {
           initialXDomain={this.props.initialXDomain}
           initialYDomain={this.props.initialYDomain}
           isRangeSelection={this.props.mouseTool === MOUSE_TOOL_SELECT}
-          leftWidth={this.leftWidth}
+          isRisShowGlobalMousePosition={this.props.isShowGlobalMousePosition}
```
pkerpedjiev: what is the `Ris` in `isRisShowGlobalMousePosition`?

app/scripts/HiGlassComponent.js
```diff
@@ -3255,16 +3296,28 @@ class HiGlassComponent extends React.Component {
...
       dataY,
-      isFrom2dTrack: hoveredTrack && hoveredTrack.is2d,
-      isFromVerticalTrack: hoveredTrack && hoveredTrack.flipText,
+      isFrom2dTrack: !!(hoveredTrack && hoveredTrack.is2d),
+      isFromVerticalTrack: !!(hoveredTrack && hoveredTrack.flipText),
       track: hoveredTrack,
       origEvt: e,
       sourceUid: this.uid,
       hoveredTracks,
+      noHoveredTracks: hoveredTracks.length === 0,
```
pkerpedjiev: See my comment on `noHoveredTracks` further down. I think it's unnecessary because we can just use `hoveredTracks.length === 0` which is just as clear and doesn't require having another variable to keep track of.

app/scripts/HiGlassComponent.js
```diff
@@ -3255,16 +3296,28 @@ class HiGlassComponent extends React.Component {
       relTrackY: (hoveredTrack && hoveredTrack.flipText) ? relTrackPos[0] : relTrackPos[1],
       dataX,
       dataY,
-      isFrom2dTrack: hoveredTrack && hoveredTrack.is2d,
-      isFromVerticalTrack: hoveredTrack && hoveredTrack.flipText,
+      isFrom2dTrack: !!(hoveredTrack && hoveredTrack.is2d),
```
pkerpedjiev: Same comment about `noHoveredTracks` below. Derived boolean variables just add clutter, IMO.

app/scripts/TiledPlot.js
flekschas: Ohh shoooot. Good catch! This is derived from a merge conflict I believe

app/scripts/utils/show-mouse-position.js
flekschas: I agree but the reason why I had to change this is because `hoveredTracks` is no longer available. This due to the fact that `hoveredTracks` is referencing track objects which are not available in another window. Therefore, the globally published `higlass.mouseMove` event does not have the property `hoveredTracks` while the locally does. An alternative would be to always publish an empty array globally but I think this would give the illusion that one could somehow get an event with a non-empty `hoveredTracks` property which is impossible. Does that make sense? 🤔🙂

app/scripts/HiGlassComponent.js
flekschas: See my comment above. I believe it's necessary because `hoveredTracks` is removed from global mouseMove events.

app/scripts/HiGlassComponent.js
flekschas: Same as above. If `hoveredTrack` was available in another window we wouldn't need it but that array is not available so we require those derived boolean variables.

PR #566, opened 2019-02-18
HiGlass docker docs updates and some minor bugs

pkerpedjiev:
## Description

Changed the higlass docker docs to point to the new higlass repo on dockerhub rather than the old gehlenborglab.

I inadvertently also included a couple bug fixes for:

* Loading matrices with resolutions (typo in the code)
* Loading local tracks such as top axis.

Why is it necessary?

Fixes #___

## Checklist

- [o] Unit tests added or updated
- [o] Updated CHANGELOG.md


app/scripts/HorizontalMultivecTrack.js
```diff
@@ -27,7 +27,7 @@ export default class HorizontalMultivecTrack extends HeatmapTiledPixiTrack {
     ctx.fillRect(0, 0, canvas.width, canvas.height);
 
     if (pixData.length) {
-      const pix = new ImageData(pixData, pixData.length / 4, canvas.height);
+      const pix = new ImageData(pixData, canvas.width, canvas.height);
```
flekschas: Could you elaborate in the PR's description what this changes? Seems to be more than just docs updates?

app/scripts/HorizontalMultivecTrack.js
```diff
@@ -78,8 +78,7 @@ export default class HorizontalMultivecTrack extends HeatmapTiledPixiTrack {
 
     this.zoomLevel = this.calculateZoomLevel();
 
-
-    if (this.tilesetInfo.resolution) {
+    if (this.tilesetInfo.resolutions) {
```
flekschas: Is this an accidental change or related to a bug?

app/scripts/TilesetFinder.js
```diff
@@ -20,7 +20,12 @@ class TilesetFinder extends React.Component {
 
     // local tracks are ones that don't have a filetype associated with them
     this.localTracks = TRACKS_INFO
-      .filter(x => x.local && !x.hidden);
+      .filter(x => x.local && !x.hidden)
+      .map((x) => {
+        const y = Object.assign({}, x);
+        y.datatype = x.datatype[0];
+        return y;
```
flekschas: Could you elaborate in the PR's description what this changes? Seems to be more than just docs updates?

app/scripts/HorizontalMultivecTrack.js
pkerpedjiev: Shit, I have no idea. I completely forgot what these changes were for. I must have stacked this PR on top of another one.

app/scripts/HorizontalMultivecTrack.js
pkerpedjiev: This is necessary because some tiles may not be full but the texture should always be the size of the canvas.

pkerpedjiev: (This is also the result of a stacked PR)

app/scripts/HorizontalMultivecTrack.js
pkerpedjiev: Oh, so this was just a typo: `this.tilesetInfo.resolution` should be `this.tilesetInfo.resolutions`.

app/scripts/TilesetFinder.js
pkerpedjiev: In the `tracks-info.js` file, datatypes are arrays because a track can be used to view multiple different datatypes. For displaying the available track types, a tileset listed in the tileset finder needs to have just a single datatype. For local track types (e.g. top axis and left axis), the track is the tileset so we have to convert the datatype array to a single value.

pkerpedjiev:
This is old, removing.

PR #572, opened 2019-02-25
Flekschas/scrollable container

### LINKED ISSUE #571: Support view scrolling for large stacks of 1D tracks
Currently, HiGlass creates a canvas element that's as high as the sum of all tracks. While this works fine in most uses cases where the height is smaller than the browser height or not that much higher. But in cases where many tracks are stacked on top of each other, the graphics card will start complaining or the performance will go down as a lot of pixels are rendered that are actually off screen.

Therefore, it would be great to have a scrolling mode (in extension to `options.bounded`) in which HiGlass creates a canvas element that is as high as the parent element in the DOM and implements a scroll behavior as shown here: https://codepen.io/wiledal/pen/ZYoNEa?editors=0010 or https://gist.github.com/slashman/79f4441a27367b661bdb86702493e44b

One thing to keep in mind is that scrolling over a track will normally lead to a zoom behavior of that track so the user either needs to disable zooming or we could utilize the ALT key to restore the normal scrolling. Similar as the ALT key can be used to get the normal context menu instead of the HiGlass context menu.

flekschas:
## Description

1. Add support for scrolling in HiGlass via the new `scrollable` option.
2. Add `option(key, value)` to the JS API for getting and setting root options.

> Why is it necessary?

1. Fixes #571 
2. Currently, options can't be changed dynamically

> How does this look like?

Like this:

Oct-10-2019 18-12-57

The key aspect here is not visible, which is that the canvas object is never larger than the gray box. Instead, within the canvas object the, y-offset of the pixi stage (which is what HiGlass is rendering on) is being adjusted according to the y-offset of the inner div element.

> How do we avoid zooming/scrolling clashes? Don't they both use the scroll event?

Excellent question! In order to activate scrollable views you have to set `scrollable` to `true` in the root option object. E.g.:

```javascript
hgApi = hglib.viewer(domEl, viewConf, { scrollable: true });
```

You can also use the newly added API endpoint called `option(key, value)` to activate/deactivate the option like so:

```javascript
hgApi.option('scrollable', true);
```

To come back to the zoom/scroll clash: when you enable scrollable, HiGlass will automatically enable zoomFixed for all views to avoid simultaneous scrolling and zooming.

> Tell me more about this fancy new JS API endpoint

This endpoint acts as a getter and setter. To get the current value of an option do:

```javascript
hgApi.option('bounded');
```

To set an option pass a value as the second argument like so:

```javascript
hgApi.option('scrollable', true);
```

Currently, all options can be get but only `scrollable` can be set.

## Checklist

- [x] Updated CHANGELOG.md


pkerpedjiev:
@flekschas - thoughts on my previous comment?

flekschas:
Sorry I completely missed this. Good question.

Not sure what the best solution would be. It'd be easy to activate classic scrolling only when holding down ALT. For now, we could also leave the decision to the app developer and have a JS API endpoint for switching between the behavior as I imagine only developer would use this functionality.

test/APITests.js
```diff
@@ -309,6 +310,64 @@ describe('API Tests', () => {
...
+
+      expect(api.option('scrollable')).toEqual(true);
+
+      const hgc = api.getComponent();
+
+      waitForTilesLoaded(hgc, () => {
+        const trackRenderer = document.querySelector('.track-renderer-events');
+        trackRenderer.scrollTop = 20;
+
+        setTimeout(() => {
```
pkerpedjiev: Why setTimeout? Is this so that the even queue has a chance to process all pending events before you check for the correct values?

app/scripts/HiGlassComponent.js
```diff
@@ -2461,7 +2479,11 @@ class HiGlassComponent extends React.Component {
     totalTrackHeight += MARGIN_HEIGHT;
     const rowHeight = this.state.rowHeight + MARGIN_HEIGHT;
 
-    if (!this.props.options.bounded) {
+    if (this.props.options.scrollable) {
```
pkerpedjiev: Comment explaining why this is necessary would be nice.

app/scripts/HiGlassComponent.js
```diff
@@ -188,8 +188,16 @@ class HiGlassComponent extends React.Component {
       setTileProxyAuthHeader(props.options.authToken);
     }
 
+    this.pixiRoot = new PIXI.Container();
+    this.pixiRoot.interactive = true;
```
pkerpedjiev: Does pixiStage also need to be interactive if `pixiRoot` is interactive?

Come to think of it, why do either of them have to be interactive? Do we use PIXI events anywhere?

flekschas: I believe both have to be interactive. Maybe the root container does not. I didn't notice any performance impact but shall I test if `this.pixiRoot` can also be non-interactive?

We use PIXI events for certain annotation tracks which support click event callbacks.

app/scripts/HiGlassComponent.js
flekschas: The difference boils down to the fact that normally `totalTrackHeight` is the same as the container height but when the user enables scrolling mode `totalTrackHeight` could be larger or smaller than the container.

I'll add a comment

test/APITests.js
flekschas: Good catch. I previously used an event to trigger scrolling until I realized that's not how the scroll event works. But then I forgot to remove the timeout again. I just removed it.

PR #579, opened 2019-03-05
Flekschas/min height bar track

flekschas:
## Description

Adds two track options to BarTrack:

- **barMinHeightAtZero**: - Height in pixel of fake bars that are to be drawn at position where the value is zero if the height is greater than zero. Drawing something at positions with zero value can help to demarcate tracks. The resulting effect is similar to the plot produced by ``horizontal-line``, i.e., even if the value is zero a thin line is drawn.

- **barMinHeightAtZeroOpacity**: - The opacity of bars drawn at positions where the value is zero. This option is useful to deemphasize the bars.

> Why is it necessary?

Having multiple tacked sparse ChIP-seq tracks can be visually a bit confusing since peaks appear as unconnected spikes. Having a connecting line helps to differentiate tracks from one another

**Without**:

screen shot 2019-03-05 at 10 59 22 am

**With 2px (1px seems to work best)**:

screen shot 2019-03-05 at 11 00 10 am

## Checklist

- [x] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## v1.5.0
 
 - Add a new option to tracks that support axis: `axisMargin` to add some margin to an axis. See [docs/examples/viewconfs/axis-margin.json](docs/examples/viewconfs/axis-margin.json) for an example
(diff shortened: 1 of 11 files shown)


flekschas:
Superseded by #580

PR #580, opened 2019-03-09
Flekschas/bar track demarcation

flekschas:
## Description

Adds two track options to BarTrack:

- **demarcationLine**: If ``true`` draw a demarcation line at the bottom of a bar track.

- **demarcationLineColor**: The color of the demarcation line. If ``undefined`` the bar fill color (``barFillColor``) will be used.

- **demarcationLineOpacity**: The opacity of the demarcation line. If ``undefined`` the bar opacity (``barOpacity``) will be used.

**With**:

Screen Shot 2019-03-09 at 12 57 22 PM

**Without**:

Screen Shot 2019-03-09 at 12 57 52 PM

## Checklist

- [x] Updated CHANGELOG.md
- [x] SVG rendering


pkerpedjiev:
Can we call this `zeroLine` or something. `demarcationLine` makes me wonder what it's demarcating. Also, I don't think we need a separate option for whether it's visible or not. We can just opacity to 0. If you really think there should be another option indicating that it's visible, can you make `zeroLineVisible`?

PR #590, opened 2019-03-16
Return promise as an API

pkerpedjiev:
## Description

Added an option to allow returning the hglib as a promise.

Why is it necessary?

The API isn't guaranteed to be returned synchronously so we need a promise to ensure that it is.

## Checklist

- [ ] Updated CHANGELOG.md


app/scripts/hglib.js
```diff
@@ -31,21 +31,42 @@ const launch = (element, config, options) => {
    * @param {Object} options Options that affect how the component is drawn and
    * and behaves.
    *
+   * @param {Object} componentCallback A callback to be called once the component
+   * is rendered. If it's not passed in, the block until the component is created.
+   *
+   * @param {Object} promise Return the ref as a promise or not.
```
flekschas: Where are those parameters? I only see `refAsPromise` but no `componentCallback` or `promise `

app/scripts/hglib.js
```diff
@@ -31,21 +31,42 @@ const launch = (element, config, options) => {
...
+        <HiGlassComponent
+          ref={c => resolve(c)}
+          options={options || {}}
+          viewConfig={config}
+        />,
+        element
+      );
+    });
+
+    return p;
```
flekschas: Why assigning the promise first? Can't it just be returned immediately?

app/scripts/hglib.js
```diff
@@ -31,21 +31,42 @@ const launch = (element, config, options) => {
...
+  }
+
+  let component = null;
 
   ReactDOM.render(
     <HiGlassComponent
-      ref={(c) => { component = c; }}
+      ref={c => component = c}
```
flekschas: Just curious, did ESLint suggest that? Usually, arow functions that do not return anything or where the return value is ignored one uses `() => {}` to show that there's no return value. Anything not critical at all. :)

app/scripts/hglib.js
```diff
@@ -91,14 +112,19 @@ const launch = (element, config, options) => {
  *
  * @return  {Object}  Newly created HiGlass component.
  */
-export const viewer = (element, viewConfig, options) => {
+export const viewer = (element, viewConfig, options,
+  refAsPromise = false) => {
```
flekschas:
```suggestion
  asPromise = false) => {
```

As a user, I have no idea what `ref` refers to. `asPromise` is shorter and to the point in my mind :)

app/scripts/hglib.js
```diff
@@ -91,14 +112,19 @@ const launch = (element, config, options) => {
...
   /**
    * Available options:
    *
    *  bounded: [true/false]
    *      Fit the container to the bounds of the element
    */
-  const hg = launch(element, viewConfig, options);
 
-  return hg.api;
+  if (refAsPromise) {
```
flekschas:
```suggestion
  if (asPromise) {
```

app/scripts/hglib.js
```diff
@@ -91,14 +112,19 @@ const launch = (element, config, options) => {
...
+    const p = launch(element, viewConfig, options, refAsPromise);
```
flekschas:
```suggestion
    return launch(element, viewConfig, options, asPromise).then(hg => hg.api);
```

Let's same some of these invaluable bytes 🤓

app/scripts/hglib.js
```diff
@@ -91,14 +112,19 @@ const launch = (element, config, options) => {
...
+    return p.then(hg => hg.api);
```
flekschas:
```suggestion
```

app/scripts/hglib.js
pkerpedjiev: Oh, I must have forgotten to take out componentCallback. I'll do that right now.

app/scripts/hglib.js
pkerpedjiev: Not sure. I'll change it back.

app/scripts/hglib.js
```diff
@@ -31,21 +31,37 @@ const launch = (element, config, options) => {
...
+      );
+    });
+  }
+
+  let component = null;
 
   ReactDOM.render(
     <HiGlassComponent
-      ref={(c) => { component = c; }}
+      ref={c => component = c}
```
pkerpedjiev:
```suggestion
      ref={(c) => { component = c; }}
```

app/scripts/hglib.js
pkerpedjiev: You're right. It can. Changed.

app/scripts/hglib.js
pkerpedjiev: Yup, you're right. Changed.

app/scripts/hglib.js
```diff
@@ -91,14 +107,18 @@ const launch = (element, config, options) => {
  *
  * @return  {Object}  Newly created HiGlass component.
  */
-export const viewer = (element, viewConfig, options) => {
+export const viewer = (element, viewConfig, options,
```
flekschas:
```suggestion
export const viewer = (element, viewConfig, options, asPromise = false) => {
```

app/scripts/hglib.js
```diff
@@ -91,14 +107,18 @@ const launch = (element, config, options) => {
...
+  asPromise = false) => {
```
flekschas:
```suggestion
```

test/APITests.js
```diff
@@ -69,6 +71,20 @@ describe('API Tests', () => {
...
+    it('creates an API using a promise', (done) => {
+      div = global.document.createElement('div');
+      global.document.body.appendChild(div);
+
+      div.setAttribute('style', 'width:600px; height: 400px; background-color: lightgreen');
+
+      const p = viewer(div, simpleCenterViewConfig, undefined, true);
+      p.then((hgApi) => {
+        expect(hgApi).not.toEqual(null);
+        api = hgApi;
```
flekschas: Is this line needed for anything? If so please ignore my suggestion. Otherwise, I think we can remove it.
```suggestion
```

mccalluc:
@pkerpedjiev : ping. As I understand it, getting the API object has been buggy because of race conditions, but this would fix it... Would you be able to resolve the conflicts and merge?

(For context, I am trying to use HiGlass to render imagery behind another transparent component, such that the `x, y, width, height` props set the initial view, and updates to these props invoke the API. My last PR #699, attempts to do this, but I believe because of the asynchronous nature of the API, it would not work either.)

manzt:
With the changes staged for v2, this will always return a `Promise` from `hg.viewer`.

PR #595, opened 2019-03-25
Minor bug fixes

pkerpedjiev:
## Description

1. Proper pubSub is used on the drag handler.
2. Value scale locks are reset when a viewconf without value scale locks is added.
3. Added a default track type from the chromsizes datatype.

Why is it necessary?

Fixes #___

## Checklist

- [x] Updated CHANGELOG.md


app/scripts/HiGlassComponent.js
```diff
@@ -130,12 +130,12 @@ class HiGlassComponent extends React.Component {
     // between views
     this.zoomLocks = {};
     this.locationLocks = {};
+    this.valueScaleLocks = {};
 
     this.prevAuthToken = props.options.authToken;
 
     // locks that keep the value scales synchronized between
     // *tracks* (which can be in different views)
-    this.valueScaleLocks = {};
```
flekschas: Should the comment be moved/removed as well?

pkerpedjiev: Yup. Moved.

PR #598, opened 2019-03-29
Overlay track fix and use default height for new tracks

pkerpedjiev:
## Description

Fixed overlay track display and used track's default height when creating new tracks.

Why is it necessary?

Overlay tracks were broken and newly created gene annotation tracks were too small to include the labels.

Fixes #___

## Checklist

- [x] Updated CHANGELOG.md


app/scripts/TiledPlot.js
```diff
@@ -1030,7 +1030,7 @@ class TiledPlot extends React.Component {
             width: this.leftWidth + this.centerWidth + this.rightWidth,
             height: this.topHeight + this.centerHeight
             + this.bottomHeight
-            + 2 * this.props.verticalMargin,
+            + this.props.paddingTop + this.props.paddingBottom,
```
flekschas: This is going to break things. Margin and padding are not the same. What is the reason for changing this?

flekschas:
```suggestion
            + this.props.marginTop + this.props.marginBottom,
```

PR #602, opened 2019-04-01
Specify available track types in `showAvailableTrackPositions`

pkerpedjiev:
## Description

When one drags a dataset over HiGlass, region outlines will pop up showing where that dataset can be displayed as a track. This relies on a JSON structure passed into `api.showAvailableTrackPositions`. Previously, we would look at the datatype in that JSON structure and determine which track types were available. Now we add the extra ability to specify which track we would expect to be created and where.

Example input JSON to `api.showAvailableTrackPositions`:

```
        {
          server: 'http://higlass.io/api/v1',
          tilesetUid: 'WtBJUYawQzS9M2WVIIHnlA',
          datatype: 'linear-labels',
          defaultTracks: ['linear-labels'],
        }
```
Also accidentally lumped in with this PR are some housekeeping files containing example commands for building the docs. Those are convenience documentation files and will likely be made obsolete in the very near future.

## Screenshot

image

## Checklist

- [x] Updated CHANGELOG.md


app/scripts/api.js
```diff
@@ -278,9 +278,26 @@ const createApi = function api(context, pubSub) {
...
+       * Show overlays where this track can be positioned. This
+       * function will take a track definition and display red
+       * or green overlays highlighting where the track can be
+       * placed on the view.
        *
        * @param {obj} track { server, tilesetUid, datatype }
+       *
+       * @example
+       *
+       *  var lineTrack = {
```
flekschas:
```suggestion
       *  let lineTrack = {
```

Let's use ES6 code as a good example 😎

docs/examples/others/test_drag.html
```diff
@@ -0,0 +1,238 @@
...
+    //     'chr1',
+    //     0,
+    //     89758810
+    //   );
+    // }, 11000);
+
+
+</script>
+<script>
+  (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
```
flekschas: Can we remove the GA code from examples?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "higlass",
-  "version": "1.5.1",
+  "version": "1.5.2",
```
flekschas: Shouldn't this be v1.5.3?

pkerpedjiev: That will sort itself out when merged with master :-)

pkerpedjiev:
> Just minor stuff but what I don't get in general is why does the user need to tell HiGlass first where he/she can drop specific datasets? If I, as a user, have to specify that first then I already know where to drop stuff and don't need visual feedback.

> I guess what I am asking is: can't the track position <-> data type be predefined so I, as a user, don't have to explicitly call the API first?

Good point. I changed the implementation so that you only have to specify the default tracks and the positions will automatically be inferred:

```
{
          server: 'http://higlass.io/api/v1',
          tilesetUid: 'WtBJUYawQzS9M2WVIIHnlA',
          datatype: 'linear-labels',
          defaultTracks: ['linear-labels'],
        }
```

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## v1.5.3
```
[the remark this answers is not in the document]
pkerpedjiev: no, I just merged with a develop branch that wasn't up to date so it's missing that part of the CHANGELOG. That will be sorted out when it's merged back.

docs/examples/others/test_drag.html
pkerpedjiev: Yup, done!

PR #605, opened 2019-04-05
Fix the error thrown by horizontal-vector-heatmap

pkerpedjiev:
## Description

`horizontal-vector_heatmap` tracks were throwing an error when used with `vector` data from `bigwig` and `bigbed` data sources. This was because they were returning data smaller than the expected tile size. This fix padds their return data to make sure it matches the specified `tile_size`.

Fixes #___

## Checklist

- [x] Updated CHANGELOG.md


app/scripts/HeatmapTiledPixiTrack.js
```diff
@@ -1026,6 +1026,19 @@ class HeatmapTiledPixiTrack extends TiledPixiTrack {
 
     this.renderingTiles.add(tile.tileId);
 
+    if (this.tilesetInfo.tile_size) {
+      if (tile.tileData.dense.length < this.tilesetInfo.tile_size) {
```
flekschas: Does that mean the server can send corrupt data? I like this fix but shouldn't we make sure the server sends proper data too? (Maybe there's another fix?)

test/HorizontalMultivecTests.js
```diff
@@ -0,0 +1,269 @@
...
+      done,
+      {
+        style: 'width:800px; height:400px; background-color: lightgreen',
+        bounded: true,
+      })
+    );
+  });
+
+  it('not have errors in the loaded viewconf', (done) => {
+    done();
```
flekschas: Where is the test? Am I missing something?

app/scripts/HeatmapTiledPixiTrack.js
pkerpedjiev: It's not exactly corrupt. It's just not complete. I think it's fine if the server sends a truncated array if the remainder is supposed to be NaN.

test/HorizontalMultivecTests.js
pkerpedjiev: The test is just loading that viewconf without throwing an error.

pkerpedjiev: Is there a way to explicitly test for no error?

Issue #490, opened 2018-12-19
2D Chromosome grid is drawn over the colormap and matrix label


Chadius:
#### Steps to reproduce
1. Create a view config with a center track. The track's contents contain a heatmap and a chromosome size layer. It doesn't matter if you add the chromosome track before or after adding the heatmap.
2. View the view config in the Higlass Display.

#### Observed behavior
- Before adding the chromosome grid:
![image 4](https://user-images.githubusercontent.com/1376213/50245589-e13c8000-03a0-11e9-84c2-c208c71551a0.png)

- After adding the grid, it obscures the heatmap. 
![image 3](https://user-images.githubusercontent.com/1376213/50245584-df72bc80-03a0-11e9-8cb9-193568ee128f.png)

- If you zoom in, the grid isn't a problem, as it fits inbetween the gaps.
![image 2](https://user-images.githubusercontent.com/1376213/50245574-dd106280-03a0-11e9-80bd-eccb7bf64f53.png)

#### Expected behavior
- The chromosome grid is drawn underneath the heatmap

Chadius:
Could you draw the grid underneath the color scale? If you see the second image, the grid is drawn on top of the scale, which it shouldn't. I'll make a mockup.

Chadius:
![clean-overlay](https://user-images.githubusercontent.com/1376213/51276320-61ef9b80-19a2-11e9-8c09-4b181c86ed46.png)
As an outcome, I'd like to at least see the scale unobstructed. Compare the upper right corner of this mockup to the second picture from the top, where the grid lines are obstructing the scale.

Chadius:
Yes, exactly. Sorry about mixing the images up when I posted initially.

flekschas:
Cool and no worries! And thanks for clarifying :) This is clearly an issue. I'll change the title of the issue slightly and, hopefully, we get around implementing it sometime soon.

pkerpedjiev-zymergen:
This is going to be tricky to implement. The problem is that the heatmap and its color map are on one layer and the grid lines are on another. There's a few options for what we can do:

1. Add an option to display grid lines natively in the heatmap track. That chromosome information is available in the cooler file and we can pass it over in the tileset info.
2. Create an option to hide the actual heatmap and only display the colorscale. That we can add two heatmaps, hide the color scale in the first and the heatmap in the second and sandwich the grid lines in between.

The first is probably slightly more elegant but the second is easier to implement.

flekschas:
[...]

PR #613, opened 2019-04-11
Mccalluc/subviewer

mccalluc:
Treat this less as something that actually should be merged, and more as a basis for questions.

Use case: I want use HiGlass literally behind the scene in Vitessce. Right now, the background imagery is just an SVG containing an `image`: As deck.gl in the foreground reports drag events, the viewBox on the svg is updated. -> I was hoping to use HiGlass instead, except with `zoomTo` instead of tweaking svg props.

1) Is there any good reason to ever to use a 0 animation time in zoomTo? I'm changing position 10 times per second: with no animation, it was not keeping up, but even an animation time of 1 looks great... If there's no good reason to use animation time of 0, would we like to make the default 1, instead?

2) Do you think I could implement the whole colorbar as SVG, and not just the sliders? As I would be using it, it would be covered by other layers. If it's not something higlass provides, then I might end up reimplementing it in Vitessce, which seems like a waste.

3) Do you have any interest in adding a `trackViewer` like this to the API? It's trivial for me to implement it on the Vitessce side, but maybe it would make this kind of use easier in the future?


pkerpedjiev:
> Is there any good reason to ever to use a 0 animation time in zoomTo? I'm changing position 10 times per second: with no animation, it was not keeping up, but even an animation time of 1 looks great... If there's no good reason to use animation time of 0, would we like to make the default 1, instead?

0 should be an immediate transition but it sounds like 0 is actually slower for you than 1? If that's the case, I'm fine with changing it to 1.

> Do you think I could implement the whole colorbar as SVG, and not just the sliders? As I would be using it, it would be covered by other layers. If it's not something higlass provides, then I might end up reimplementing it in Vitessce, which seems like a waste.

I think we may have originally had that. The main reason it's rendered on the canvas instead of SVG is that it's easier to export as PNG and it stays in the same position as the heatmap. If it's rendered as SVG, it'll be rendered above all other canvas-based tracks. Neither solution is ideal since if we render on the canvas layer, other tracks can occlude the colorscale. See this, for example: https://github.com/higlass/higlass/issues/490

Other suggestions for how to handle this are welcome. And we should actually have an SVG implementation in `exportAsSVG`.

> Do you have any interest in adding a trackViewer like this to the API? It's trivial for me to implement it on the Vitessce side, but maybe it would make this kind of use easier in the future?

Yes, definitely. Maybe eventually we can break that out into its own library to keep the size down.

app/scripts/hglib.js
```diff
@@ -102,3 +102,53 @@ export const viewer = (element, viewConfig, options) => {
 
   return hg.api;
 };
+
+export const trackViewer = (element, [xMin, xMax, yMin, yMax], trackConfig, options) => {
+  if (!trackConfig.options.colorbarPosition) {
+    trackConfig.options.colorbarPosition = 'hidden';
+    // If the colorbar were all SVG, I could change the colorbar Z position,
+    // and move the whole thing around, but apart from the handlebars,
+    // I think it's all part of the canvas, right?
```
pkerpedjiev: Yes. See here: https://github.com/higlass/higlass/blob/fdd8f08c3769849a129f5f98c7d6ffd8a3106dd8/app/scripts/HeatmapTiledPixiTrack.js#L507

app/scripts/hglib.js
```diff
@@ -102,3 +102,53 @@ export const viewer = (element, viewConfig, options) => {
...
+  }
+  if (!trackConfig.options.labelTextOpacity) {
+    trackConfig.options.labelTextOpacity = 0;
+    // TODO: There is still a rectangle in the lower right... Is there an easy way to just turn
```
pkerpedjiev: You should be able to set `labelPosition` to `hidden`. You could also set `labelBackgroundOpacity` to 0.

mccalluc:
@pkerpedjiev : When #626 is merged, I would like to rebase this, see it work locally, and then merge it. Are there any issues here which you would consider blockers?

pkerpedjiev:
#626 is merged. I updated your PR to include a docstring. Could you go over it and see if it makes sense and agrees with your intention? I also removed the unused options parameter.

Why are you returning {id, hgApi} instead of just hgApi? If you're setting id to the same thing all the time, why not just say what it is in the docstring and not return it?

mccalluc:
> Why are you returning {id, hgApi} instead of just hgApi? If you're setting id to the same thing all the time, why not just say what it is in the docstring and not return it?

I think I framed this as a don't-repeat-yourself issue, and as a maintainability issue.
- DRY: It's better to have the literal value in one place than three (where it's used internally, where it's documented, and where it would be used externally).
- Maintainability: If I'm looking at my code in the future, and there's some magic string in it, I may wonder where it came from. Hopefully all three instances of the value will match, but if not...

I'll look at the other changes now.

mccalluc:
@pkerpedjiev : I think this is ready to go now. Regarding the `id` that is returned, I could change it. I like the way it is, but I could be looking at the problem wrong.

I've regenerated `examples.html` to include the example I just added: It had been pretty-printed at somepoint, and I could add that to the script, if it's useful to anyone... but since I've thought of it as something that would just be regenerated periodically, I don't see it as a priority.

mccalluc:
@pkerpedjiev : ping!

PR #619, opened 2019-04-16
animateTime=0 stutters, at least for me, and =1 works well

mccalluc:
## Description

Give a default 1ms animation time
Follow-up to https://github.com/higlass/higlass/pull/613#issuecomment-483063637

Why is it necessary?

With `animationTime=0`, it seems to be redrawing from scratch each time, and couldn't keep up with refreshes faster than once per second, or so.

## Checklist

- [x] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## in progress
+
+- Improve zoomTo performance by specifying non-zero animationTime.
+
 ## v1.5.4
 
 - Fix the multiple component passive event issue by replacing the dom-event.js handlers with a class so that each component maintains its own context
--- a/app/scripts/api.js
+++ b/app/scripts/api.js
@@ -395,7 +395,7 @@ const createApi = function api(context, pubSub) {
         end1Abs,
         start2Abs,
         end2Abs,
-        animateTime = 0,
+        animateTime = 1,
       ) {
         self.zoomTo(viewUid, start1Abs, end1Abs, start2Abs, end2Abs, animateTime);
       },


mccalluc:
`development` has been failing linting: `import/no-cycle`... I'll see if I can understand what introduced the problem, or if it's an unpinned dependency issue...

flekschas:
> With `animationTime=0,` it seems to be redrawing from scratch each time, and couldn't keep up with refreshes faster than once per second, or so.

Do you have a screenshot demonstrating the problem and the fix? It'd be interesting to know when and why `setZoom` fails.

@pkerpedjiev `animationTime=0` was to introduced to avoid 2 parameters (https://github.com/higlass/higlass/commit/592dbc44410edff4f7df903f162d2e0483b67977#diff-a1c9615d4ea6647745b9fd979eee0f4dL838).

Setting `animationTime` to `1` by default essentially will always trigger the whole zoom machinery https://github.com/higlass/higlass/blob/develop/app/scripts/TrackRenderer.js#L1003-L1026 for 1 millisecond.

mccalluc:
@flekschas : Looking at this again, the difference is really between animateTime=0 (which is fine) and animateTime=null (which stutters). I think at some point I must have left animateTime off, assumed that was the same as zero, and noted that animateTime=1 looked good, without ever actually checking 0.

To reproduce,
```
git checkout  mccalluc/subviewer
npm run start
```
and compare http://localhost:8081/others/trackViewer.html with http://localhost:8081/others/trackViewer.html?animateTime=0

I'll close this PR, but if the difference between null and 0 surprises you, open a new issue? Thanks!

PR #629, opened 2019-04-21
Destroy texture-containing elements

pkerpedjiev:
## Description

Destroy sprites when necessary so that they're not maintained in PIXI's texture cache.

Why is it necessary?

Slows down memory leaks.

Fixes #___

## Checklist

- [x] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## v1.5.5
+
+- Destroy heatmap sprites and axis texts to mitigate memory leak
+
 ## v1.5.4
 
 - Fix the multiple component passive event issue by replacing the dom-event.js handlers with a class so that each component maintains its own context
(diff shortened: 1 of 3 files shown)


flekschas:
I did a quick test and it looks like the general memory leak is fixed. GPU memory went from 750 MB to 1GB at most and stabilizes around 950 MB.

I am not sure if this is just my subjective perception but it feels like panning and zooming got clunkier, i.e., the UI freezes more often for a few milliseconds. Presumably because rendering and freeing the memory is done synchronously? Maybe I can take a look of texture clearing can be outsourced to a worker? (For another PR)

pkerpedjiev:
Yeah, that could be. It could also be that because we're freeing memory we weren't previously the garbage collector gets invoked more often. Still, probably better a little clunkier than leaky.

A big PR would be to outsource all the texture creation and removal to a worker thread. I've started that twice before and always gave up but there's no good reason why it can't be done. Maybe a task for a lazy rainy weekend :-)

PR #662, opened 2019-05-23
Flekschas/awesome lazy share view as link modal

### LINKED ISSUE #647: Exported view link is incorrectly displayed when exportViewUrl points elsewhere
_Commit_: v1.5.6

#### Steps to reproduce
1. In your local viewconfig set `exportViewUrl: 'http://higlass.io/api/v1/viewconfs',`
2. Export the view as a link

#### Observed behavior
- The exported view link will begin with `http://localhost:8080`

#### Expected behavior
- The exported view link should begin with `http://higlass.io`

### LINKED ISSUE #648: Add a copy-to-clipboard button to the export viewconf URL dialog
Because highlighting text and ctrl+c is just so laborious... ;)

Also, the viewconf URLs can be longer than the dialog box and it's easy to miss a few characters at the end by not dragging the cursor far enough... I've also found it easy to accidentally click somewhere outside the modal, which closes it before I manage to copy (that's probably a bug).

flekschas:
## Description

- Shows correct URL when specifying an absolute URL as `exportViewUrl` in the viewconf.
- When exporting a view as a link, the link will automatically be selected and copied to the clipboard. Also to make it explicit, I added a button to copy the link explicitly.

> Why is it necessary?

Fixes #647 
Fixes #648

## Checklist

- [x] Updated CHANGELOG.md


app/scripts/ExportLinkDialog.js
```diff
@@ -1,26 +1,50 @@
...
+          Copied to clipboard!
+          <span aria-label="Hooray" role="img"> 🎉</span>
+        </p>
+        <div styleName="export-link-dialog-wrapper">
+          <input
+            ref={(element) => {
+              this.input = element;
+              element.focus();
+              element.select();
+              document.execCommand('copy');
```
pkerpedjiev: Is there a precedent for a UI automatically placing something in the clipboard? I'm concerned that it might be annoying if unexpected.

flekschas: I was asking myself the same question but isn't the whole purpose of the link export modal to present you a URL that you copy?

flekschas: @nvictus what do you think? Since you started the call for more laziness in regards to the modal 😁

nvictus: How would anyone know of this auto-clipboarding side effect? I think a simple copy-to-clipboard button is something people are very accustomed to:

image

nvictus: I can confirm, though, that the modal currently closes when clicking anywhere outside of it, which isn't typically how modals work. It makes it easy to accidentally lose the modal before getting a chance to copy the link.

That's probably the real source of frustration.

flekschas:
It’s written above the link that is was auto-copied. So it should be kind of obvious. Anyway, I will remove it.

Closing a dialog on outer click is actually quite often implemented. I’m annoyed when the only way to close a modal is via the little “X” button.

pkerpedjiev:
I'm OK with closing a dialog by clicking outside it. Did you remove the auto-copy? Let's merge this in after that's done.

flekschas:
Not yet. Will try to do today or next week

Issue #652, opened 2019-05-17
PNG Promise does not appear to resolve (or reject)


alexpreynolds:
_Commit_: 

HiGlass v1.5.3

#### Steps to reproduce
```
        let pngPromise = this.hgView.api.exportAsPngBlobPromise();
        console.log("pngPromise", pngPromise);
        pngPromise
          .then((blob) => {
            let reader = new FileReader(); 
            reader.addEventListener("loadend", function() {
              let array = new Uint8Array(reader.result.slice(0,8));
              /* do stuff with array */
            }); 
            reader.readAsArrayBuffer(blob);
          })
          .catch((err) => {
            throw Error(err);
          })
          .finally(function() {
            alert("PNG export attempt finished")
          });
```

#### Observed behavior
- The Promise `pngPromise` remains in `pending` mode and does not seem to resolve. Via console.log:

```
pngPromise 
  Promise {<pending>}
    __proto__: Promise
      [[PromiseStatus]]: "pending"
      [[PromiseValue]]: undefined
```

#### Expected behavior
- The Promise either resolves or rejects

flekschas:
I am unable to reproduce this in `v1.5.7`. When I do:

```js
setTimeout(() => {
  hgApi.exportAsPngBlobPromise()
    .then((blob) => {
      const reader = new FileReader();
      reader.addEventListener('loadend', () => {
        const array = new Uint8Array(reader.result.slice(0, 8));
        console.log('all good', array);
      });
      reader.readAsArrayBuffer(blob);
    })
    .catch((err) => {
      throw Error(err);
    })
    .finally(() => {
      console.log('PNG export attempt finished');
    });
}, 2500);
```

I'll see:

```
HiGlass v1.5.7: http://higlass.io
PNG export attempt finished
all good Uint8Array(8) [137, 80, 78, 71, 13, 10, 26, 10]
```

Could you provide a complete example with the issue?

alexpreynolds:
I upgraded HiGlass to v1.5.7 since filing this issue.

Here is the code that I have, relevant to PNG export (which fails) and SVG export (which works):

```
  onClickExportItemSelect = (name) => {
    switch (name) {
      ...
      case "png":
        console.log("this.hgView.api", this.hgView.api);
        let pngPromise = this.hgView.api.exportAsPngBlobPromise();
        console.log("pngPromise", pngPromise);
        pngPromise
          .then((blob) => {
            console.log(blob);
            let reader = new FileReader(); 
[...]

PR #663, opened 2019-05-23
Flekschas/export rendered view as png

flekschas:
## Description

When exporting a view as PNG, the actual rendered view is now exported. Previously the SVG rendering was exported as a PNG but the SVG export isn't 100% the same as the rendered view in the browser.

> Why is it necessary?

To simplify the export and export the actually visible views

## Checklist

- [x] Updated CHANGELOG.md


pkerpedjiev:
Ohhhh... the reason it was exporting the SVG output is that we can have SVG tracks that are rendered independent of PIXI.

Probably better to keep the SVG-based export. Sorry for not realizing that earlier,

flekschas:
Oh shoot that's true. Too bad because it looked like this PR fixed #652. Maybe there's another way to just render the current SVG on top to avoid having to re-export everything as SVG.

pkerpedjiev:
Is it possible to fix #652 without jumping to using the rendered view to to export?

flekschas:
Probably but I don’t know how from the top of my head :)

PR #674, opened 2019-06-04
Mccalluc/stop exporting svgtrack

### LINKED ISSUE #444: Why is SVGTrack publicly exported?
_Commit_: develop

`hglib.js` exports `SVGTrack` publicly. What's the purpose? Is this only for testing purposes? If so I'd suggest having a separate file that exports everything needed for testing (maybe the file could be called `hglib.testing.js`). Not that people start to depend on it.

```javascript
export { default as SVGTrack } from './SVGTrack';
```

mccalluc:
## Description

- Stop exporting SVGTrack

If there is a reason to keep this, great, we just don't know what it is.

Fixes #444

## Checklist

- [ ] Updated CHANGELOG.md


pkerpedjiev:
I suspect that there is a reason it was exported but we can remove it and see if anything breaks.

PR #680, opened 2019-06-07
Mccalluc/generated index.json

mccalluc:
## Description

- Pull list of fixtures from another json file.
- Confirm that it's up to date

Why is it necessary?
- We want to make sure we run the tests against an up-to-date set of files.

Fixes #___

## Checklist

- [ ] Updated CHANGELOG.md


travis_build_test.sh
```diff
@@ -11,6 +11,13 @@ start eslint
 ./node_modules/eslint/bin/eslint.js karma.conf.js app test
 end eslint
 
+start examples
+REGEN_SCRIPT='./docs/examples/regenerate-index.py'
+INDEX_JSON='docs/examples/index.json'
+diff --ignore-space-change <( $REGEN_SCRIPT ) $INDEX_JSON \
```
pkerpedjiev: Sorry, I'm having a hard time wrapping my head around what this is doing. Could you explain it in a sentence or two? Or perhaps add a comment?

pkerpedjiev: It looks like it doesn't actually update index.json but just warns the user if it needs to be updated?

mccalluc: Yep, that's what it does: I'll add a comment. The `<( )` syntax is for process substitution: The output of the command is treated like a file argument that can be read from. (There's also `>( )` for writing.)

mccalluc: Merged recent changes from and develop, and fixed this. I'll go ahead and merge if it comes back green.

PR #682, opened 2019-06-08
Mccalluc/even more fussy validation

mccalluc:
## Description

- Tighter schema

You'll probably want changes here... A descriptive schema (that wouldn't report errors with any of the working viewconfs that are out there) will be different than a normative schema (that describes a state we think we can get to)... or something in between.

## Checklist

- [ ] Updated CHANGELOG.md

### DIFF
--- a/app/schema.json
+++ b/app/schema.json
@@ -27,8 +27,8 @@
       "additionalProperties": false,
       "required": ["locksByViewUid", "locksDict"],
       "properties": {
-        "locksByViewUid": { "type": "object", "additionalProperties": false },
-        "locksDict": { "type": "object", "additionalProperties": false }
+        "locksByViewUid": { "type": "object" },
+        "locksDict": { "type": "object" }
       }
     },
 
@@ -53,7 +53,7 @@
     "genomePositionSearchBox": {
       "type": "object",
       "additionalProperties": false,
-      "required": ["autocompleteServer", "autocompleteId", "chromInfoServer", "chromInfoId", "visible"],
+      "required": ["chromInfoServer", "chromInfoId", "visible"],
       "properties": {
         "autocompleteServer": { "type": "string" },
         "autocompleteId": { "type": "string" },
--- a/app/schema.json
+++ b/app/schema.json
@@ -45,8 +45,33 @@
         "genomePositionSearchBox": { "type": "object", "TODO": "more precise" },
         "chromInfoPath": { "type": "string" },
         "tracks": { "$ref": "#/definitions/tracks_object" },
-        "layout": { "type": "object", "TODO": "more precise" },
-        "overlays": { "type": "array" }
+        "layout": { "$ref": "#/definitions/layout" },
+        "overlays": { "type": "array", "items": { "$ref": "#/definitions/overlay" } }
+      }
+    },
+
+    "overlay": {
+      "type": "object",
+      "properties": {
+        "uid": { "type": "string" },
+        "type": { "type": "string" },
+        "includes": { "type": "array" },
+        "options": { "type": "object" }
+      }
+    },
+
+    "layout": {
+      "type": "object",
+      "additionalProperties": false,
+      "required": ["w", "h", "x", "y", "moved", "static"],
+      "properties": {
+        "w": { "type": "number" },
+        "h": { "type": "number" },
+        "x": { "type": "number" },
+        "y": { "type": "number" },
+        "i": { "type": "string" },
+        "moved": { "type": "boolean" },
+        "static": { "type": "boolean" }
       }
     },
 
@@ -65,7 +90,7 @@
     },
 
     "tracks_array": {
-      "note": "We can't enforce this because every track type has a different model. :(",
+      "note": "I don't think we can enforce this because every track type has a different model?",
       "type": "array",
       "items": {
         "type": "object",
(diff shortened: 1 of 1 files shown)


app/schema.json
```diff
@@ -27,8 +27,8 @@
       "additionalProperties": false,
       "required": ["locksByViewUid", "locksDict"],
       "properties": {
-        "locksByViewUid": { "type": "object", "TODO": "more precise" },
-        "locksDict": { "type": "object", "TODO": "more precise" }
+        "locksByViewUid": { "type": "object", "additionalProperties": false },
```
pkerpedjiev: Shouldn't there be properties listed here?

Sorry, my knowledge of JSON schemas is rather limited.

app/schema.json
```diff
@@ -42,11 +42,92 @@
...
+        "tracks": { "$ref": "#/definitions/tracks_object" },
+        "layout": { "$ref": "#/definitions/layout" },
+        "overlays": { "type": "array", "items": { "$ref": "#/definitions/overlay" } }
+      }
+    },
+
+    "genomePositionSearchBox": {
+      "type": "object",
+      "additionalProperties": false,
+      "required": ["autocompleteServer", "autocompleteId", "chromInfoServer", "chromInfoId", "visible"],
```
pkerpedjiev: autoCompleteServer and autoCompleteId shouldn't be required. You can have a Genome Position Search Box with just chromsome info.

app/schema.json
```diff
@@ -42,11 +42,92 @@
...
+        "top": { "type": "array" },
+        "bottom": { "type": "array" },
+        "center": { "type": "array" },
+        "whole": { "type": "array" },
+        "gallery": { "type": "array" }
+      }
+    },
+
+    "tracks_array": {
+      "note": "I don't think we can enforce this because every track type has a different model?",
```
pkerpedjiev: Yeah, this one is tricky. There's a lot of variation in terms of how a track can be defined. This looks like a good start though.

app/schema.json
mccalluc: Actually, I should just delete the `additionalProperties`: This is a map from UIDs to locks, so the property names can't be validated... but we could make assertions about the locks.

app/schema.json
mccalluc: Just took them out of the required list. Thanks.

app/schema.json
mccalluc: Would you like this much to be enforced? This would pass with all the examples that are checked in, but it's not enforced: In the section above, we only say that `left`, `right`, etc are arrays: 
```
"tracks_object": {
      "type": "object",
      "additionalProperties": false,
      "properties": {
        "left": { "type": "array" },
        "right": { "type": "array" },
        "top": { "type": "array" },
        "bottom": { "type": "array" },
        "center": { "type": "array" },
        "whole": { "type": "array" },
        "gallery": { "type": "array" }
      }
    },
```
Options:
- Status quo
- Validate against it. ( `"type": "array"` -> `"$ref": "#/definitions/tracks_array"` )
- Even stricter: We could check the track types, and make sure that those tracks have the particular properties that belong.

mccalluc:
@pkerpedjiev : Made a couple small tweaks. I'm happy to work more on this, really... but I'm not sure where the point of diminishing returns is, and that's where your experience would be really useful. One possible route:
- Check in this much as-is.
- If you wanted to provide a good example of viewconf that demonstrated locks, I could write schema against that.
- Is the validation of overlays sufficient?
- If there are particular track types that get more use, or whose configuration is more error prone, I should write out special cases just for them... but I don't think we want to try to try to handle every track type?

pkerpedjiev:
I'm not sure where the point of diminishing returns is either but there's two things that would make this easier to update:

1. Add a line to the pull request template asking contributors to make sure that they update the validation schema when changing the viewconf.
2. Update the developer docs (https://github.com/higlass/higlass/blob/develop/docs/developer.rst) with a section describing how to run the validation locally from the command line.

Is there some way to warn about sections that are not validated? I really appreciate you working on this and I think the greatest impact would be to gently shepherd us toward incremental improvements as we continue development.

To reply to your specific questions:


> Check in this much as-is.
Yes please.

 > If you wanted to provide a good example of viewconf that demonstrated locks, I could write schema against that.

Here's one: https://higlass.io/app/?config=CaySmEh3TsqChLBrek9JTw

This has all of location, zoom and value scale locks.

> Is the validation of overlays sufficient?

At a first glance, yes, but to be perfectly honest, I haven't looked at them in a while so take my judgement with a grain of salt.

> If there are particular track types that get more use, or whose configuration is more error prone, I should write out special cases just for them... but I don't think we want to try to handle every track type?

Heatmap tracks are probably the most used. In an ideal world, we would take the available track configuration options from here: https://github.com/higlass/higlass/blob/develop/app/scripts/configs/tracks-info.js and use that in the validation. In the real world, that's difficult because a viewconf can contain custom tracks that are defined in outside repos (e.g. https://github.com/higlass/higlass-multivec)

mccalluc:
Thank you! I'll merge this, and either take the next steps, or file issues based on your comments for someone to follow-up later.

mccalluc:
> two things that would make this easier to update

PR made.

> Is there some way to warn about sections that are not validated? 

I think the biggest gap is the tracks, and the best thing is just to keep them in mind. Beyond that, I think what makes the most sense is just to keep the schema pretty strict (required properties, etc) so there will be some warnings... hopefully not every time, but enough to serve as a reminder.

> Check in this much as-is.
> Yes please.

done.

> Here's one

Thanks -- will make another PR.

> Heatmap tracks are probably the most used

I'll add just that much then, as an example of how others could be validated.

PR #686, opened 2019-06-14
Mccalluc/locks schema

mccalluc:
## Description

- Validate locks
- Namespace the definitions to keep things neater
- Enumerate all the track types
- For `combined` and `heatmap`, suggest how a more fine-grained validation could be done.

I think this is enough for now. I think it would be good to articulate a policy moving forward: If you do see a validation warning, should you
- file a bug?
- expand the schema to handle that case?
- simplify the output so it conforms to schema?

You might also consider having layers of validation, manipulating this data structure before loading it into ajv... for example, if you remove enum_track, you get tight validation on only a few tracks... or you might change it to a catch-all, if you only want to catch larger structural problems, but don't want to validate individual tracks.

I think this will be all for me here, and I should get back to the single-track react component.

## Checklist

- [ ] Updated CHANGELOG.md

### DIFF
--- a/app/schema.json
+++ b/app/schema.json
@@ -38,104 +38,106 @@
       "required": ["tracks", "layout"],
       "properties": {
         "uid": { "type": "string" },
-        "initialXDomain": { "$ref": "#/definitions/domain" },
-        "initialYDomain": { "$ref": "#/definitions/domain" },
+        "initialXDomain": { "$ref": "#/definitions/viewProperties/domain" },
+        "initialYDomain": { "$ref": "#/definitions/viewProperties/domain" },
         "autocompleteSource": { "type": "string" },
         "genomePositionSearchBoxVisible": { "type": "boolean" },
-        "genomePositionSearchBox": { "$ref": "#/definitions/genomePositionSearchBox" },
+        "genomePositionSearchBox": { "$ref": "#/definitions/viewProperties/genomePositionSearchBox" },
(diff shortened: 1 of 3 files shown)


app/schema.json
```diff
@@ -4,31 +4,105 @@
...
+  "required": [],
   "properties": {
     "editable": { "type": "boolean" },
     "zoomFixed": { "type": "boolean" },
     "exportViewUrl": { "type": "string" },
     "trackSourceServers": {
       "type": "array",
-      "items": { "type": "string" }
+      "items": { "type": "string" },
+      "minLength": 1
```
pkerpedjiev: You don't necessarily need to have a track source server. You could have it empty if you don't want to allow people to add or remove tracks.

pkerpedjiev: But then again, I suppose if it's going to be empty, you might as well not include the section at all.

app/schema.json
```diff
@@ -4,31 +4,105 @@
...
   "definitions": {
 
     "locks": {
-      "type": "object",
-      "additionalProperties": false,
-      "required": ["locksByViewUid", "locksDict"],
-      "properties": {
-        "locksByViewUid": { "type": "object" },
-        "locksDict": { "type": "object" }
+      "string22": {
```
pkerpedjiev: This isn't necessary. I've been using 22 character slugid's but as long as the identifiers are unique, there's no reason why they need to be constrained to 22 character strings.

PR #687, opened 2019-06-14
Flekschas/upper right lower left heatmap

flekschas:
## Description

I added an option to define the _extent_ of a heatmap. The extent can be one of the following options:
- full (default)
- upper-right
- lower-left

**For example**, Rao 2014, GM12878 vs k562
Screen Shot 2019-06-13 at 7 01 04 PM

The UI allows setting the extent through the context menu.

Here's an example of how to set up the triangular split matrix:

When you add a second heatmap,
1. HiGlass will automatically set the background of that heatmap to `transparent`
2. HiGlass will copy the `showMousePosition`, `mousePositionColor`, and `showTooltip` options from the previous heatmap (previous in the z-level sense)

If you change the `extent` of a central matrix to `upper-right` or `lower-left` HiGlass will automatically change the position of the colormap to `topRight` and `bottomLeft` for the `name` and `colorbar` respectively.

Jun-26-2019 16-39-37

Finally, I added an option to add x-offset. I use this x-offset for the is used when the `name` and `colorbar` position is identical. Right now there's a clash and both overlay. With this PR the label will be pushed _inwards_ by the width of the colorbar.

> Why is it necessary?

To compare 2 heatmaps in one central plot

## Checklist

- [x] Updated CHANGELOG.md


flekschas:
> Would this not be simpler to implement with a PIXI mask instead of going through and setting values to NaN? So the view gets rendered as it was before and if someone sets an extent, we just add a PIXI mask that covers the relevant triangle?

Maybe easier but most certainly it's a huge waste of resources. We duplicate the data in the first place to get from a triangular matrix to a rectangular matrix. Why not just simple *not* duplicate the data to get the original triangular matrix? I am setting values to `NaN` after copying them from the lower left to the upper right triangle.

> What's the mouseOver behavior with this solution? If I do a mouseOver part of the track that isn't in the visible extent, will that be passed down to the track below?

Yes right now the way mouse over works it shows both tracks but only one of them will have a value while the other one has no value. 

> Does SVG export work?

Not right now, there's a bug in the code that I have to fix firsy

> Why no tests?

The PR is not ready yet, which is why I haven't added anyone for review ;) I like to set up a PR so future additions are visible even if the code isn't ready to be merged yet.

There are a couple more things to get right before this can be merged (apart from adding docs/test):

- [ ] Fix the tooltip
- [ ] Update and restrict the position of the label and colormap changer
- [ ] Update and restrict the background color to transparent

flekschas:
@pete As a matter of facts, the SVG might work (or not) but there's a general bug breaking the examples right now. See #690

pete:
@flekschas I'm glad you think SVG might work; I like SVG in general, except that the `<path>` element always felt a little like they should have just started with that and only did that, but I forgave them when I realized that the `d` attribute is basically the LOGO turtle.

Though I do always try to respond when people ask for my input by @'ing me on Github, I am afraid that my own time constraints prohibit expending the necessary effort to give an informed opinion on #690, so I hope you will excuse me from that. :)

flekschas:
@pete Oh nooo, I did it again 😅 Sorry for the mentioning you again I need to be more careful and type more slowly when it comes to GitHub mentionings as I tend to mix them up with Slack usernames. Anyway, thanks for the kind reply :)

pkerpedjiev:
All sounds good. Let me know when you'd like me to take a look again.

app/scripts/HiGlassComponent.js
```diff
@@ -2332,17 +2333,26 @@ class HiGlassComponent extends React.Component {
         // if it's a combined track, we just need to add this track to the
         // contents
         tracks.center[0].contents.push(newTrack);
+
+        if (newTrack.type === 'heatmap') {
+          this.compatibilityfyStackedHeatmaps(newTrack, tracks.center[0]);
```
pkerpedjiev: Why you doing this? Comment please.

pkerpedjiev: Oh, I see in the function definition below. A short comment would still be nice :-)

flekschas: Sure how about the comment below?
```suggestion
          // For stacked heatmaps we will adjust some options automatically for convenience
          this.compatibilityfyStackedHeatmaps(newTrack, tracks.center[0]);
```

flekschas:
Yes, you are right and generally speaking it already works when the extent of the upper heatmap is set. The reason why I also set it for the lower heatmap is that this will automatically position the colorbar and name label.

I'll add a test tomorrow

flekschas:
@pkerpedjiev I added tests and even more convenientification: when you have 2 stacked heatmaps and change the extent of one of them the extent of the other will be changed accordingly. See the gif below.

Jun-27-2019 13-26-41

PR #693, opened 2019-06-18
Mccalluc/track component

mccalluc:
## Description

- Add `HiGlassTrackComponent` to API
- Add demo
- The demo did not work initially: I think when the call to `launch` is itself wrapped inside React, the order of execution change, and the component ref was not being set before exit... So I explicitly call createRef outside to avoid this reordering.
- I had a test, but it didn't fail, even when the example failed, but I could try again.

Why is it necessary?
- I want to be able to use HG to display a single track with React: The ref was being set only after `launch` returned. 

## Checklist

- [ ] Updated CHANGELOG.md


app/scripts/hglib.js
```diff
@@ -148,8 +147,54 @@ export const trackViewer = (element, [xMin, xMax, yMin, yMax], trackConfig) => {
       },
     ],
   };
-  return {
-    id,
-    hgApi: viewer(element, viewConfig, { bounded: true })
-  };
+  const hgApi = viewer(element, viewConfig, { bounded: true });
+  console.warn('>>> hgApi in trackViewer?', hgApi);
```
pkerpedjiev: Should this be here?

app/scripts/hglib.js
```diff
@@ -148,8 +147,54 @@ export const trackViewer = (element, [xMin, xMax, yMin, yMax], trackConfig) => {
...
+  }
+
+  initTrackViewer() {
+    const {
+      trackConfig,
+      x, y, width, height,
+    } = this.props;
+    const element = document.getElementById(this.id);
+    const { id, hgApi } = trackViewer(element, [x, x + width, y, y + height], trackConfig);
+    console.warn('>>> hgApi in HiGlassTrackComponent?', hgApi);
```
pkerpedjiev: Should this be here?

app/scripts/hglib.js
```diff
@@ -148,8 +147,54 @@ export const trackViewer = (element, [xMin, xMax, yMin, yMax], trackConfig) => {
...
+    this.viewUid = id;
+    this.viewer = hgApi;
+  }
+
+  zoomTo(x, y, width, height) {
```
pkerpedjiev: Would you mind adding a docstring? Is x,y the upper left, center?

app/scripts/hglib.js
```diff
@@ -148,8 +147,54 @@ export const trackViewer = (element, [xMin, xMax, yMin, yMax], trackConfig) => {
...
+  console.warn('>>> hgApi in trackViewer?', hgApi);
+  return { id, hgApi };
 };
+
+/**
+ * React component wrapper around trackViewer.
+ * Accepts x, y, width, and height props, in addition to trackConfig,
+ * so HiGlass can be used to provide background imagery for Deck.gl.
+ */
+export class HiGlassTrackComponent extends React.Component {
```
pkerpedjiev: This is should maybe be in its own file?

app/scripts/hglib.js
mccalluc: nope. Fixed.

app/scripts/hglib.js
mccalluc: nope. Fixed.

app/scripts/hglib.js
mccalluc: yep. Fixed.

app/scripts/hglib.js
mccalluc: done.

mccalluc:
I think the crux of this is 
```diff
-  let component;
-
+  const ref = React.createRef();
   ReactDOM.render(
     <HiGlassComponent
-      ref={(c) => { component = c; }}
+      ref={ref}
       options={options || {}}
       viewConfig={config}
     />,
     element,
   );
-  return component;
+  return ref;
```

We can't assume that the ref function will have been evaluated when `render` returns... and although this change works for me, there are a lot of tests failing now. I'll look into them now, but feedback would be welcome... Is there a better idiom for what we were trying to do here? Or should I take different approach in my new single-track component?

mccalluc:
yay! I'll merge this, but I do want to improve testing.

mccalluc:
D'oh: I don't think this actually fixed anything: 
```
HiGlassTrackComponent.componentDidMount() ->
   .initTrackViewer() ->
trackViewer() ->
viewer() ->
launch() ->
ReactDOM.render(<HiGlassComponent ...>)
```
The `HiGlassComponent.render` is still not being called before we ask for the api object.

PR #694, opened 2019-06-18
Flekschas/throttle debounce tile fetching

flekschas:
## Description

Use a combination of throttling and debouncing for fetching new tiles.

> Why is it necessary?

Before new tiles were only fetched 100 ms after the _last_ call. Normally this is fine but for animations like `hgApi.zoomTo(...)` calls will come in every few milliseconds and so the fetch call is delayed until the entire animation is over. This can be super annoying because the animation will look like it's going nowhere.

With throttling we will ensure that at least every 100 ms (defined in `config/index.js`) a new fetch request is triggered, which ensure that during many high-frequent zoom calls new tiles are fetched and rendered.

For example, with debouncing only a `hgApi.zoomTo(...)` can lead to this result:

debounce-only

While with throttling + debouncing the new tiles will already be loaded:

throttle-and-debounce

A real video would be cooler but the gifs are too 

## Checklist

- [x] Updated CHANGELOG.md


app/scripts/services/tile-proxy.js
```diff
@@ -83,19 +84,28 @@ const debounce = (func, wait) => {
     requestMapper = {};
   };
 
-  const debounced = (request, ...args) => {
-    bundleRequests(request);
+  const callFunc = (request, ...args) => {
```
pkerpedjiev: Can we get a comment describing what this is for?

pkerpedjiev: Oh, I see that it's used in debounced. It would still be helpful to have a short comment to that effect.

app/scripts/utils/throttle-and-debounce.js
```diff
@@ -0,0 +1,52 @@
...
+    clearTimeout(timeout);
+    reset();
+  };
+
+  debounced.immediate = (...args) => {
+    func(...args);
+  };
+
+  let wait = false;
+  const throttled = (request, ...args) => {
```
pkerpedjiev: It would be nice to have a short explanation of why we're using throttling and how it works together with debouncing. I know you wrote this in the PR but if someone is looking at the code they won't have access to the PR.

flekschas: Good point. I wanted to do that but somehow forgot. Will add an explanation.

app/scripts/services/tile-proxy.js
flekschas: Sure will do!

flekschas: Done!

app/scripts/utils/throttle-and-debounce.js
flekschas: Done!

PR #700, opened 2019-06-20
Flekschas/cleanup viewconfig

flekschas:
## Description

Remove unused the following properties from the viewconfig:

- `genomePositionSearchBoxVisible` (this property is part of `genomePositionSearchBox`)
- `view.layout.i` (autopopulated by HiGlass)
- `view.layout.moved` (autopopulated by react-grid-layout)
- `view.layout.static` (autopopulated by react-grid-layout)
- `track.position` (was used internally only and actually not needed at all)
- `track.name` (`name` is defined on `track.options`)

Expanded the view config schema.

Also, I fixed 2 bugs I found along the way.

Also, I refactored how we _walk_ over visible tracks using _visitor-like-pattern_. This should (a) be more resource friendly and (b) avoid the need for manipulating the track definitions.

> Why is it necessary?

We want a clean view config for SciPy (and in general :D)

## Checklist

- [x] Updated CHANGELOG.md


app/schema.json
```diff
@@ -123,14 +133,48 @@
     "viewProperties": {
       "genomePositionSearchBox": {
         "type": "object",
-        "additionalProperties": false,
-        "required": ["chromInfoServer", "chromInfoId", "visible"],
+        "title": "The Genome Position Search Box Schema",
+        "required": [
+          "autocompleteServer",
```
pkerpedjiev: The autocomplete stuff isn't actually required. You can have a genome position search box just chromInfo.

app/scripts/HiGlassComponent.js
```diff
@@ -3209,31 +3187,23 @@ class HiGlassComponent extends React.Component {
         this.yScales[v.uid] = scaleLinear().domain(v.initialYDomain);
       }
 
-      // Add names to all the tracks
-      let looseTracks = positionedTracksToAllTracks(v.tracks);
-
-      // give tracks their default names (e.g. 'type': 'top-axis'
-      // will get a name of 'Top Axis'
-      looseTracks = this.addUidsToTracks(looseTracks);
```
pkerpedjiev: Are you sure this can be removed? Doesn't this add uids to tracks that don't have them?

app/scripts/HiGlassComponent.js
```diff
@@ -3209,31 +3187,23 @@ class HiGlassComponent extends React.Component {
...
-      looseTracks = this.addNamesToTracks(looseTracks);
```
pkerpedjiev: Are you sure this can be removed? The comment seems to suggest that this gives tracks a default name like "Top Axis". Otherwise they may be displayed as "undefined" when a label is selected.

app/schema.json
flekschas: Must have been a merge issue. Thanks for spotting!
```suggestion
```

app/scripts/HiGlassComponent.js
flekschas: Yes because `TiledPlot` does the exact same thing again. Look at line 81.

Or am I missing something?

app/scripts/HiGlassComponent.js
flekschas: Yes, this function is useless as far as I can tell. First, it overwrites any existing name and second, it adds the name to the track root object instead of the `options` property so it's anyway not used.

This used to be the function for reference:

```
const config = this.getTrackInfo(track.type);
if (config && config.name) track.name = config.name;
```

As far as I can tell `track.name` isn't used anywhere or do I overlooking something?

app/schema.json
```diff
@@ -123,14 +133,47 @@
     "viewProperties": {
       "genomePositionSearchBox": {
         "type": "object",
-        "additionalProperties": false,
-        "required": ["chromInfoServer", "chromInfoId", "visible"],
+        "title": "The Genome Position Search Box Schema",
+        "required": [
+          "autocompleteId",
```
pkerpedjiev: This isn't required.

app/scripts/HiGlassComponent.js
pkerpedjiev: Yeah, I suppose you're right. Maybe just quickly check to make sure that the Top Axis track still has a label without this function?

app/schema.json
flekschas: Fixed. Thanks for finding!

app/scripts/utils/visit-positioned-tracks.js
```diff
@@ -0,0 +1,19 @@
+import visitTracks from './visit-tracks';
+
+const visitPositionedTracks = (
+  positionedTracks,
+  visitor,
+  inclCombinedTracks = true
+) => {
```
pkerpedjiev: Would you mind adding a short docstring? `inclCombinedTracks` is a little difficult for me to understand. If it's false, all subchildren of the combined tracks are skipped? Why would it ever be false?

PR #701, opened 2019-06-21
Mccalluc/fix json schema validation in tests

mccalluc:
## Description

- Running the tests, I noticed that there were new json validation warnings. Tracking that down, I realized we had viewconfs checked in, beyond the ones under docs. If we store all our viewconf fixtures the same way, it will make it easier to scan them all.

Why is it necessary?

- It's not, but it's moving in a direction where validating the fixtures as part of the tests becomes much simpler, though for the moment this doesn't validate any more of the viewconfs.

## Checklist

- [ ] Updated CHANGELOG.md


mccalluc:
Started work on #702: Now that we have all the viewconfs as json, we can use commandline tools to quickly validate them.

travis_build_test.sh
```diff
@@ -11,15 +11,11 @@ start eslint
...
-REGEN_SCRIPT='./docs/examples/regenerate-index.py'
-INDEX_JSON='docs/examples/index.json'
-diff --ignore-space-change <( $REGEN_SCRIPT ) $INDEX_JSON \
-  || die "Update fixture index: '$REGEN_SCRIPT > $INDEX_JSON'"
-# Compare the output of the regen script to the current index...
-# If there's a difference, it will be shown,
-# along with a one-liner to fix the problem.
-end examples
+start viewconfs
+CMD='./docs/examples/regenerate-index.py docs/examples/viewconfs test/view-configs test/view-configs-more'
```
pkerpedjiev: Comment please? What is this doing and why?

docs/examples/regenerate-index.py
```diff
@@ -2,14 +2,19 @@
...
 
-def main():
-    dir = os.path.dirname(os.path.realpath(__file__))
-    viewconfs_dir = os.path.join(dir, 'viewconfs')
-    print(json.dumps(
-        sorted(os.listdir(viewconfs_dir)),
-        indent=2
-    ))
+def main(dirs):
+    not_dirs = [f for f in dirs if not os.path.isdir(f)]
```
pkerpedjiev: Would you mind adding a docstring and an error message in case the correct arguments aren't passed? This would be helpful in documenting what exactly this function does and how it can be used. I like `argparse` but there may be other packages that can do this.

pkerpedjiev:
Sorry, I'm not entirely clear what's happening here. There's an `_index.json`, there's `test/view-confs`, `test/view-confs-more` and it looks like `index.json` is being generated by `regenerate-index.py`. What's the relationship between these different pieces?

travis_build_test.sh
mccalluc: It was regenerating lists of the files in each of these directories, so the JS tests could iterate over all.... but it's simpler just to get rid of the JS tests which can't list a directory.

docs/examples/regenerate-index.py
mccalluc: got rid of this python script

mccalluc:
I hope the picture is more clear in #705. Closing.

PR #703, opened 2019-06-22
Add an option to make 2D annotations non-interactive

flekschas:
## Description

Added an option called `interactive` for `2d-annotation` tracks, which by default is `true` (same behavior as before this PR. Setting `interactive: false` will draw all annotation bounding boxes onto the same graphics object and dramatically improve the performance at the expense of not having interactivity anymore, i.e., the user can't click on and select individual annotations anymore.

The option is accessible through the track context menu.

> Why is it necessary?

This greatly improves the performance when rendering ~1000 annotations from 30-40 FPS to 60 FPS

## Checklist

- [ ] Add test
- [x] Updated CHANGELOG.md

### DIFF
--- a/app/scripts/Annotations2dTrack.js
+++ b/app/scripts/Annotations2dTrack.js
@@ -197,6 +197,23 @@ class Annotations2dTrack extends TiledPixiTrack {
 
     if (!tile.tileData.length) return;
 
+    if (this.options.interactive === false) {
+      graphics.clear();
+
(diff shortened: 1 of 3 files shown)


manzt:
We're ramping up work for a higlass v2 release and closing stale PRs to reduce noise. This has been open for over a year, so I'm closing it for now. Feel free to reopen if still relevant. Thanks!

PR #702, opened 2019-06-21
Mccalluc/validate everything

mccalluc:
## Description

What was changed in this pull request?
- Building on #701 demonstrate how all the fixtures could be validated.
- In progress

Why is it necessary?

Fixes #___

## Checklist

- [ ] Unit tests added or updated
- [ ] Documentation added or updated
- [ ] Example added or updated
- [ ] Update schema.json if there are changes to the viewconf JSON structure format
- [ ] Screenshot for visual changes (e.g. new tracks)
- [ ] Updated CHANGELOG.md

PR #705, opened 2019-06-24
Mccalluc/validate everything

mccalluc:
## Description

- Combines #701 and #702, and addresses Pete's questions, and finishes what I began.
- SchemaTests.js has been removed: It required the creation and maintenance of index files, and it ran very slowly because each file was fetched via http. We lose coverage of the re-exported viewconfs, but this is much simpler, and much faster.
- Tweaked schema to accommodate more variations.

Why is it necessary?
- More consistency in how viewconf fixtures are managed.
- Tests run much faster

## Checklist

- [ ] Updated CHANGELOG.md


app/schema.json
```diff
@@ -209,14 +212,23 @@
               "children": { "type": "array" }
             }
           },
+          "created": { "type": "string" },
```
pkerpedjiev: This shouldn't be here. This is actually erroneously included in exported viewconfs and viewconfs that contain it are incorrect.

app/schema.json
```diff
@@ -209,14 +212,23 @@
...
           "server": { "type": "string" },
           "tilesetUid": { "type": "string" },
           "position": { "type": "string" },
           "options": { "type": "object" },
           "name": { "type": "string" },
           "transforms": { "type": "array" },
           "width": { "type": "number" },
-          "height": { "type": "number" }
+          "height": { "type": "number" },
+          "binsPerDimension": { "type": "number" },
```
pkerpedjiev: This is returned by the server in tilesetInfo and shouldn't be here. It's erroneously included in exported viewconfs and viewconfs that contain it are incorrect.

app/schema.json
```diff
@@ -209,14 +212,23 @@
...
+          "orientation": { "type": "string" },
+          "local": { "type": "boolean" },
+          "thumbnail": { },
+          "maxWidth": { "type": "number" },
+          "minWidth": { "type": "number" },
+          "minHeight": { "type": "number" },
+          "filetype": { "type": "string" }
```
pkerpedjiev: This is returned by the server in tilesetInfo and shouldn't be here. It's erroneously included in exported viewconfs and viewconfs that contain it are incorrect.

app/schema.json
```diff
@@ -209,14 +212,23 @@
...
           "options": { "type": "object" },
           "name": { "type": "string" },
           "transforms": { "type": "array" },
           "width": { "type": "number" },
-          "height": { "type": "number" }
+          "height": { "type": "number" },
+          "binsPerDimension": { "type": "number" },
+          "orientation": { "type": "string" },
+          "local": { "type": "boolean" },
+          "thumbnail": { },
```
pkerpedjiev: This shouldn't be here either, I don't think. @flekschas do you know if the thumbnail should be in the viewconf output?

app/schema.json
mccalluc: Would you prefer that the fixture which includes this be moved to view-configs-invalid, or that that fixture just be corrected? (I lean towards the latter.)

app/schema.json
mccalluc: Same question: I lean towards correcting these viewconf fixtures.

app/schema.json
mccalluc: Same question: I lean towards correcting these viewconf fixtures.

mccalluc:
- I'll fix the fixtures, if that sounds like the right route? Please confirm.
- On the directories:
  - The `view-configs` directory existed before this PR, except the files were js instead of json.
  - The files in `view-configs-more` all were in one huge `view-configs.js`. `view-configs.js` still exists, but it just imports and re-exports each file in `view-configs-more` now.
  - `view-configs-tracks` were also in `view-configs.js`, but judging by the names they are just tracks, so I've separated them.

I don't think there's any need for `view-configs` and `view-configs-more` to be separate, but I didn't want to change too many things at once. (The files inside do have different naming conventions: I might settle on mixedCase, so the filename can match the variable name?) (If this is even worth doing.)

pkerpedjiev:
> I'll fix the fixtures, if that sounds like the right route? Please confirm.

It does indeed!

> (The files inside do have different naming conventions: I might settle on mixedCase, so the filename can match the variable name?) (If this is even worth doing.)

Up to you. I tend to prefer dashes in filenames but don't really care all that much either way.

All the rest seems reasonable.

mccalluc:
- Fixed the lines in the schema which you had flagged.
- Tightened up a few other places in the schema. (Some cases, like thumbnail, which is now either an empty object or null are probably too tight, but I believe that means we should improve our fixtures.)
- Switch from a loop to xargs, and the test runs much faster.
- Sort long lists alphabetically.

PR #709, opened 2019-06-27
Flekschas/auto zoom to extent

### LINKED ISSUE #593: Small genome matrices are hard to display
First of all, thank you for developing HiGlass, this is an awesome tool.

One of the issue we currently have is that when loading a small genome matrix, like _Drosophila melanogaster_, the Hi-C matrix is **very** small. More often than not, we have to carefully zoom in and try to find the matrix, and usually we just "loose" the matrix (zoomed out too much, or zoomed in but not in the correct place). It is the case both when displaying a Hi-C matrix on side panels or at the center (although it is easier to find the matrix on the side panels of a window).

Wouldn't it be possible to automatically resize the window coordinates to the actual beginning and end of a matrix when it's added to the center or lateral panels?

Best,
Gautier

flekschas:
## Description

Zoom to data extent when a new track is added to an empty view when only one view exists.

> Why is it necessary?

Fixes #593

## Checklist

- [x] Updated CHANGELOG.md


app/scripts/HiGlassComponent.js
```diff
@@ -2357,6 +2360,18 @@ class HiGlassComponent extends React.Component {
 
     this.adjustLayoutToTrackSizes(this.state.views[viewId]);
 
+    if (Object.keys(this.state.views).length === 1 && numTracks === 0) {
+      // Zoom to data extent since this is the first track we added and we only
+      // have one view
+
+      // This doesn't work because the tilesetInfo is probably not there yet
+      this.handleZoomToData(viewId);
```
pkerpedjiev: haha, if this doesn't work then why keep it there?

app/scripts/TiledPlot.js
```diff
@@ -121,6 +122,11 @@ class TiledPlot extends React.Component {
       addDivisorDialog: null,
     };
 
+    // This should be `true` until one tracks was added and initialized.
```
pkerpedjiev:
```suggestion
    // This should be `true` a track is added and initialized.
```

PR #724, opened 2019-07-05
Fix #722

### LINKED ISSUE #722: Make the current resolution label optional
_Commit_: develop

#### Steps to reproduce
1. Load a vector track

#### Observed behavior
- When the track label is shown, HiGlass additionally prints `[Current data resolution: ...]`
- There is no way to just show the label without this extra information: https://github.com/higlass/higlass/blob/develop/app/scripts/PixiTrack.js#L360

Screen Shot 2019-07-05 at 3 24 42 PM

#### Expected behavior
- This is an advanced feature so it shouldn't be displayed by default.
- It should be possible to show the track label without this extra information.

I can see that this information is useful for genomic data but for general vector data is just adding clutter to the view that a normal user does not need.

flekschas:
## Description

Added `labelShowResolution` as an option to allow hiding the `[Current data resolution...]` text

> Why is it necessary?

Fixes #722 

**With data resolution:**

Screen Shot 2019-07-05 at 5 47 00 PM

**Without data resolution:**

Screen Shot 2019-07-05 at 5 46 50 PM

## Checklist

- [x] Updated CHANGELOG.md


pkerpedjiev:
Same with vertical-1d-value-interval

flekschas:
Good catch. Will remove the option.

PR #726, opened 2019-07-06
Switch to the new share link url

pkerpedjiev:
## Description

The export as link dialog will now create links with the format `server/l/?d=uuid`. This will return a small dummy page with metadata for link unfurling.

Why is it necessary?

So that we can use link unfurling in Slack and Twitter

Fixes #___

## Checklist

- [x] Updated CHANGELOG.md

### DIFF
--- a/app/scripts/HiGlassComponent.js
+++ b/app/scripts/HiGlassComponent.js
@@ -2806,7 +2806,7 @@ class HiGlassComponent extends React.Component {
       })
       .then(_json => ({
         id: _json.uid,
-        url: `${parsedUrl.origin}/app/?config=${_json.uid}`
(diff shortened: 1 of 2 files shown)


app/scripts/HiGlassComponent.js
```diff
@@ -2806,7 +2806,7 @@ class HiGlassComponent extends React.Component {
       })
       .then(_json => ({
         id: _json.uid,
-        url: `${parsedUrl.origin}/app/?config=${_json.uid}`
+        url: `${parsedUrl.origin}/l/?d=${_json.uid}`
```
flekschas: Not sure I am a fan of this URL schema. Do we have to space or why did we decide not to use something more readable?
We can push this out but I would prefer `/link/?config=`

PR #727, opened 2019-07-06
Only compare tracks that are suppose to receive a tileset info

### LINKED ISSUE #593: Small genome matrices are hard to display

flekschas:
## Description

Only compare tracks that are supposed to receive a tileset info. Previously all tracks were compared against tracks that received a tileset info. If you initialized HG with a `top-axis` and `heatmap` track `TiledPlot.tracksByUidInit`was of length 2 but only one of the tracks would ever receive a tileset info so `TiledPlot.handleZoomToData()` is never called. 

> Why is it necessary?

Truly fix #593

## Checklist

- [x] Updated CHANGELOG.md


app/scripts/TiledPlot.js
```diff
@@ -457,26 +457,33 @@ class TiledPlot extends React.Component {
     ) return;
 
     // Get the total number of track that are expecting a tilesetInfo
-    const allTilesetInfos = Object.keys(this.trackRenderer.trackDefObjects)
+    const allTracksWithTilesetInfos = Object.keys(this.trackRenderer.trackDefObjects)
       // Map track to a list of tileset infos
       .map((trackUuid) => {
         const track = this.trackRenderer.trackDefObjects[trackUuid].trackObject;
         if (track.childTracks) {
           return track.childTracks.map(childTrack => childTrack.tilesetInfo);
```
pkerpedjiev: The line below this returns a track but this returns a tilesetInfo. Is that right?

pkerpedjiev: And not just a tilesetInfo but a list of tilesetInfos

flekschas: Oh good catch

PR #731, opened 2019-07-08
Improvements to the BedLikeTrack

pkerpedjiev-zymergen:
## Description

This PR introduces a number of improvements to the bedlike track like the ability to add labels and a better rendering of the features when significantly zoomed in. It also introduces separating items according to strand and assigning each strand a separate color.

This does not affect the previous functionality of stacking by value (which assumes no strand information).

Fixes #___

## Checklist

- [x] Updated CHANGELOG.md

## Screenshots

Zoomed out, labels on:

image

Zoomed in, labels off:

image


app/scripts/utils/interval-tree.js
```diff
@@ -118,8 +118,8 @@ IntervalTree.prototype.contains = function _contains(point) {
 };
 
 function intersects(a, b) {
-  return (a[0] <= b[0] && a[1] >= b[0]) || (a[0] <= b[1] && a[1] >= b[1])
-      || (b[0] <= a[0] && b[1] >= a[0]) || (b[0] <= a[1] && b[1] >= a[1]);
```
flekschas: What's the functional difference? Are the borders excluded? Just out of curiosity I am wondering, was this a bug or is this just a different choice of behavior.

docs/examples/viewconfs/bedlike-track.json
```diff
@@ -0,0 +1,118 @@
...
+         "genomePositionSearchBox":{
+            "autocompleteServer":"http://higlass.io/api/v1",
+            "chromInfoServer":"http://higlass.io/api/v1",
+            "visible":true,
+            "chromInfoId":"hg19",
+            "autocompleteId":"OHJakQICQD6gTD7skx4EWA"
+         }
+      }
+   ],
+   "zoomLocks":{
```
flekschas: Why all this empty code?

docs/track_types.rst
```diff
@@ -45,7 +54,16 @@ Intervals can visually encode information using the following three ``options``:
...
     A tuple defining the minimum and maximum range value for color encoding.
 
-Here is an example snippet
+``plusStrandColor``
+    For stranded mode, the color of the plus strand entries
+
+``minusStrandColor``
+    For stranded mode, the color of the minus strand entries
+
+``fillColor``
```
flekschas: Wasn't fill color removed? I am confused haha

app/scripts/BedLikeTrack.js
```diff
@@ -48,57 +62,62 @@ class BedLikeTrack extends HorizontalTiled1DPixiTrack {
...
+          const text = new PIXI.Text(geneInfo[3], TEXT_STYLE);
+          if (this.flipText) { text.scale.x = -1; }
 
-        text.anchor.x = 0.5;
-        text.anchor.y = 1;
+          text.anchor.x = 0.5;
+          text.anchor.y = 0.5;
 
-        tile.texts[geneInfo[3]] = text; // index by geneName
+          tile.texts[td.uid] = text; // index by geneName
```
flekschas: Just a sanity check comment: I assume the change from `geneInfo[3]` to `td.uid` is expected? :)

app/scripts/BedLikeTrack.js
```diff
@@ -336,36 +404,109 @@ class BedLikeTrack extends HorizontalTiled1DPixiTrack {
...
         }
       }
     }
   }
 
+  renderTile(tile) {
+    let maxPlusRows = 1;
+    let maxMinusRows = 1;
+
+    for (const tile1 of this.visibleAndFetchedTiles()) {
```
flekschas: What's the difference between `tile` and `tile1`? Is `renderTile(tile)` just rendering `tile` or is it rendering all visible and fetched tiles? I am slightly confused.

flekschas: Oh is `tile1` just an already rendered tile? I guess I know what's going on. It might still be nice to rename `tile1` into something more meaningful.

app/scripts/configs/options-info.js
```diff
@@ -488,6 +496,18 @@ export const OPTIONS_INFO = {
     }
   },
 
+
+  annotationHeight: {
+    name: 'Annotation Height',
```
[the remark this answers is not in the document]
pkerpedjiev: Yes

app/scripts/configs/tracks-info.js
```diff
@@ -1702,7 +1702,8 @@ export const TRACKS_INFO = [
     name: 'BED-like track',
     thumbnail: null,
     availableOptions: [
-      'fillColor',
```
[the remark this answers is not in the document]
pkerpedjiev: My mistake.

Snippet from Slack conversation for the record:

pete 2:50 PM
On the fillColor parameter, do you have thoughts on whether to remove it? There’s currently two modes of operation:
No strand information: it doesn’t make sense to say plusStrandColor because items aren’t necessarily on a strand, therefore fillColor should be used
+ / - strand information: fillColor as a default, each strand can set using [plus/minus]StrandColor individually.
Does it make sense to keep it and use it like that?

new messages
Fritz 2:55 PM
I am more confused as to wether it's still there or not
I'd definitely keep it
What you describe above makes complete sense
I was just worried it got removed
pete 2:58 PM
:+1:
yeah, it looks like I removed it from availableOptions which was a mistake but it’s definitely still used by the renderer

app/scripts/utils/interval-tree.js
pkerpedjiev: Yeah, borders are excluded. The function difference is minor but very visible. The bedlike track stacks overlapping intervals. So imagine that you have two intervals:

```
|--|
```
and
```
    |--|
```

If we counted borders as intersects then they would be rendered like this:

```
|--|
     |--|
```

We would prefer to render them as adjacent to each other:

```
|--|--|
```

Does that make sense?

docs/track_types.rst
pkerpedjiev: No, not removed. My mistake above. Sorry about that.

app/scripts/BedLikeTrack.js
pkerpedjiev: Yup, expected. Using `geneInfo[3]` (which corresponds to gene name) can lead to lots of names not being rendered due to segments having the same name.

app/scripts/BedLikeTrack.js
pkerpedjiev: Yeah, it's just used to iterate over all the other tiles so that we can find out what the maximum number of rows is.

pkerpedjiev: I'll rename it.

app/scripts/utils/interval-tree.js
flekschas: It makes complete sense! Thanks for the explanation :) It's not necessary but maybe we can save the comment into the function's docstring so we don't forget about the rationale? It's up to you.

Issue #737, opened 2019-07-16
tile-proxy imports non-existing function


flekschas:
_Commit_: `v1.6.5`

I noticed that [`tile-proxy.js`](https://github.com/higlass/higlass/blob/develop/app/scripts/services/tile-proxy.js) [imports a function called `workerFetchTiles()`](https://github.com/higlass/higlass/blob/develop/app/scripts/services/tile-proxy.js#L5) from [`worker.js`](https://github.com/higlass/higlass/blob/develop/app/scripts/worker.js) but that function does not exist. Even more confusing the import statement is silenced to not cause ESLint to fail.

What is the intended use case? Is this just dead code or did somehow the function get lost?

I just want to avoid removing code that is supposed to work.

flekschas:
From @pkerpedjiev on Slack:

> That function is left over from when I tried moving the tiling to a web worker

Since the function doesn't even exist in a commented-out form I guess we can either remove the import and related method or at least comment it out.

PR #738, opened 2019-07-16
Flekschas/copy tile data before mutating

flekschas:
## Description

Copy tile data before mutating it.

> Why is it necessary?

To render the same tileset twice (once in the upper right triangle and once in the lower left triangle) without causing rendering issues for tiles on the diagonal.

Generally, due to throttling/debouncing tile requests are bundled (which is a great thing!) and duplicated tile requests are deduplicated. Unfortunately, this causes issues when rendering the same tileset in the upper and lower triangle because for tile mirroring we're mutating the tile data. Previously this wasn't an issue because central heatmaps with a full extent simply copied the data. For triangular matrices, we now only move the data such that one triangle of the diagonal tiles remains transparent. Therefore, I added a 2-way copying strategy. By default the base object is shallowly-copied, meaning that the dense data remains the same even if multiple tracks use the same tile data. This should be super cheap. Once a track wants to mutate the tile data it first creates a deep copy of the dense data so in case other tracks want the unmutated data they are not affected.

**Before:**

Screen Shot 2019-07-16 at 4 19 54 PM

**After (with fix):**

Screen Shot 2019-07-16 at 4 20 33 PM

## Checklist

-~[ ] Example added or updated~
- [x] Updated CHANGELOG.md

### DIFF
--- a/app/scripts/services/tile-proxy.js
+++ b/app/scripts/services/tile-proxy.js
@@ -2,7 +2,6 @@ import { range } from 'd3-array';
 import slugid from 'slugid';
 
 import {
-  // workerFetchTiles,
   workerGetTiles,
   workerSetPix,
 } from '../worker';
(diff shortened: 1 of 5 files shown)


app/scripts/services/tile-proxy.js
```diff
@@ -2,7 +2,7 @@ import { range } from 'd3-array';
 import slugid from 'slugid';
 
 import {
-  workerFetchTiles, // eslint-disable-line import/named
+  // workerFetchTiles,
```
mccalluc:
```suggestion
```

This could just be deleted: See #737

flekschas: I leave it in as commented out code because other commented-out functions rely on it. The code needs proper cleaning in another PR, just removing this line makes things even more confusing.

flekschas:
Ok I've removed all the commented out code

PR #739, opened 2019-07-16
Flekschas/do not leak dark mode

flekschas:
## Description

- Stop leaking the dark mode setting by using providers
- Added an option to set default options by theme and added default options for a couple track types
- Removed the unused `options` item from the view header context menu

> Why is it necessary?

If one of multiple HiGlass instances was initialized in dark mode, subsequent interactions with the other instances turned them into the dark mode too.

See http://localhost:8080/others/dark-theme.html for an example

Screen Shot 2019-07-16 at 7 01 57 PM

In the example above notice than the view config did not have to be changed! Unfortunately, this convenience won't work after exporting because we explicitly export default options and importing a view config doesn't specify which option was explicitly set and which one was set by default.

## Checklist

- [x] Updated CHANGELOG.md


app/scripts/configs/tracks-info.js
```diff
@@ -1509,17 +1571,24 @@ export const TRACKS_INFO = [
       'color',
       'stroke',
       'fontSize',
-      'fontIsAligned',
+      'fontIsLeftAligned',
```
pkerpedjiev: should this be `fontAlignment` rather than `fontIsLeftAligned`?

flekschas: Now I am thinking it should really be `textAlign` (to adhere to CSS) but right now it's `fontIsLeftAligned`. This change is in general unrelated to the PR but I noticed this typo some time ago and just fixed it. Oh and `fontIsAligned` does not exist so it didn't do anything before.

flekschas:
> Is the theme included in the viewconf?

No, right now it's not but this is a good point. My thinking was that the theme should have nothing to do with the views, which is true so far when the views only rely on the default colors. But after exporting the viewconfig the coloring of tracks etc. will be stored in the viewconfig.

Maybe we should think about whether the theme should be saved in the view config or not.

PR #761, opened 2019-08-17
Fix drag over coloring

pkerpedjiev:
## Description

The coloring for dragover divs had gone astray to contain only red or green. Now when someone requests drag over regions, the places where a track can be added are colored blue, places where it can't are colored red and the region where the mouse hovers are colored green.

> Why is it necessary?

To distinguish allowed regions (e.g. top, left, bottom, right for a line track) vs. disallowed (e.g. 'center' for a line track).

> Screenshots

For a line track: correct

image

vs... all red.

Fixes #___

## Checklist

- [x] Updated CHANGELOG.md


flekschas:
Why is "Set proper GitHub labels (e.g. v1.6+, ignore if you don't know)" checked but the PR doesn't have any label? 🤔 Is this a bug fix PR or an enhancement?

CHANGELOG.md
```diff
@@ -1,8 +1,14 @@
+## v1.6.6
+
+- Fixed the drag over colors
```
flekschas: Can you elaborate a bit more? I don't know what this means.

app/scripts/DragListeningDiv.js
```diff
@@ -18,7 +18,13 @@ class DragListeningDiv extends React.Component {
 
   render() {
     // color red if not enabled, green if a track is not top and red otherwise
```
flekschas: Very minor: This comment seems outdated not that there is blue.

pkerpedjiev: Good catch. Fixed.

CHANGELOG.md
pkerpedjiev: Spruced that up a bit.

pkerpedjiev:
> Why is "Set proper GitHub labels (e.g. v1.6+, ignore if you don't know)" checked but the PR doesn't have any label? 🤔 Is this a bug fix PR or an enhancement?

Sorry about that, I thought that referred to the milestones. Added a label just now.

PR #762, opened 2019-08-17
Value scale zooming

pkerpedjiev:
## Description

This PR lays the framework for zooming along the y-axis in horizontal tracks.

> Why is it necessary?

To enable zooming into individual reads once the pileup track is implemented.

Fixes #___

## Screen grab

If it were enabled, it would look like this...

Aug-17-2019 14-56-29

To properly support this we need to consider what happens when new data is received and the scales are changed. That is not addressed in this PR so the code implementing the actual zooming is commented out.

The hooks to the track methods, however, are still there and can be implemented by any other track.

Even though the actual value scale zooming is *not* implemented, this PR introduces a substantial change in behavior wherein 1D tracks no longer allow panning along the secondary axis. Horizontal tracks, for example, do not allow panning along the vertical higlass axis and vertical tracks do not allow panning along the horizontal axis. Panning along center tracks is allowed along both dimensions.

Aug-25-2019 19-57-40

## Checklist

- [x] Updated CHANGELOG.md


app/scripts/TrackRenderer.js
```diff
@@ -1033,13 +1032,44 @@ class TrackRenderer extends React.Component {
...
+  }
+
   /**
    * Respond to a zoom event.
    *
    * We need to update our local record of the zoom transform and apply it
    * to all the tracks.
    */
   zoomed() {
+    if (this.props.valueScaleZoom || this.valueScaleZooming) {
```
flekschas: Would it make sense to directly check the `event` object here to see if `altKey` or `shiftKey` is `true`? That would simplify the code a little bit and avoid state changes.

app/scripts/BarTrack.js
```diff
@@ -294,6 +297,78 @@ class BarTrack extends HorizontalLine1DPixiTrack {
...
+    // // see the reasoning behind why the code in
+    // // zoomedY is commented out.
+
+    // Object.values(this.fetchedTiles).forEach((tile) => {
+    //   const vst = this.valueScaleTransform;
+    //   const height = this.dimensions[1];
+
+    //   // clamp at the bottom and top
+    //   if (
+    //     vst.y + dY / vst.k > -(vst.k - 1) * height
```
flekschas: Would it also be okay if we destruct `vst` to `y` and `k`? When you implement `vst` makes sense but when you read this the first time it's a bit hard to follow. I am just thinking of:

```
const { y, k } = this.valueScaleTransform;
```

instead of
```
const vst = this.valueScaleTransform;
```

app/scripts/BarTrack.js
```diff
@@ -294,6 +297,78 @@ class BarTrack extends HorizontalLine1DPixiTrack {
...
+    // const height = this.dimensions[1];
+
+    // // clamp at the bottom
+    // t1 = Math.max(t1, -(k1 - 1) * height);
+
+    // // clamp at the top
+    // t1 = Math.min(t1, 0);
+
+    // // right now, the point at position 162 is at position 0
+    // // 0 = 1 * 162 - 162
```
flekschas: Is this just an example or a leftover from the development?

flekschas:
Just as docs within PR: y scales can be changed via `SHIFT` + `wheel` or dragging (is that correct?)

PR #769, opened 2019-09-07
Add api call to show track chooser

pkerpedjiev:
## Description

This PR adds an API call to let an external program show the track chooser (yellow overlay where the mouse is). It also enables the track chooser to respond to drag events in addition to mouse move events.

> Why is it necessary?

If we want to be able to drag something over a track we need a way to be able to show which track we're currently over.

## Screen capture

Sep-07-2019 14-49-22

Fixes #___

## Checklist

- [x] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Next release
+
+- Add API call for show track chooser
+
+_[Detailed changes since v1.6.8](https://github.com/higlass/higlass/compare/v1.6.8...develop)_
+
 ## v1.6.8
 
 - Add infrastructure for value scale zooming
@@ -5,7 +11,7 @@
 - Added support for value scale zooming
 - Added utils/track-utils to provide track functionality for use without inheritance
 
-_[Detailed changes since v1.6.7](https://github.com/higlass/higlass/compare/v1.6.7...develop)_
+_[Detailed changes since v1.6.7](https://github.com/higlass/higlass/compare/v1.6.7...v1.6.8)_
 
 - Implemented a genbank data fetcher for gene annotation tracks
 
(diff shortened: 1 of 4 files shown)


app/scripts/TiledPlot.js
```diff
@@ -495,19 +495,18 @@ class TiledPlot extends React.Component {
     });
   }
 
-  handleOverlayMouseLeave() {
-    this.setState({
-      mouseOverOverlayUid: null,
-    });
+  handleOverlayMouseLeave(uid) {
+    if (uid === this.state.mouseOverOverlayUid) {
```
flekschas: Why is this needed? Shouldn't it not matter which track is left?

CHANGELOG.md
```diff
@@ -1,11 +1,17 @@
+## Next release
+
+- Add API call for show track chooser
```
flekschas: Could you add a pointer what this API is called? Maybe something like `hgApi. showTrackChooser()`?

app/scripts/TiledPlot.js
pkerpedjiev: It's necessary because of the order of events. Sometimes the dragStart event begins on an element before mouseLeave is called on the leaving element and that prevents the overlay from appearing. Took a surprisingly long time to figure that out actually.

CHANGELOG.md
pkerpedjiev: Yup 👍

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
...
```
pkerpedjiev:
```suggestion
- Add API call for show track chooser (hgApi. showTrackChooser()`)
```

PR #776, opened 2019-09-15
Allow viewconf to override min dims with `minHeight` or `minWidth` props (2)

pkerpedjiev:
## Description

When a viewconf is generated programmatically, a track height or width may be less than preset defaults. This PR modifies that check to look for a `minHeight` or `minWidth` parameter set in the viewconf track, and use that to apply new bounds.

> Why is it necessary?

Helps resolve discussion on Slack (https://higlass.slack.com/archives/CCDG12ZKP/p1568182133018000)

## Checklist

- [x] Updated CHANGELOG.md


alexpreynolds:
I built from this branch and it appears to work fine — thanks for adding tests!

pkerpedjiev:
This ended up being a larger operation than initially expected but I think it's ready to merge so it'll be in the next patch release 👍 

Thanks for the getting the ball rolling!

PR #781, opened 2019-09-27
Rename unsafe react lifecycle methods and update packages

pkerpedjiev:
## Description

Rename the React lifecycle methods that are going to be deprecated according to their guidelines.
## Checklist

- [x] Updated CHANGELOG.md


pkerpedjiev:
>  One question, does anything regarding react-resizable or react-sortable change?

Yes, they have the same  problem internally so React complains about them. Updating them fixed the issue in those libraries but we still get warnings from a couple other libraries (react-grid-layout). Whenever they release new versions we can update them as well.

PR #785, opened 2019-10-03
Flekschas/fancified overlays

flekschas:
## Description

Fancified overlay borders and added support for outlines

New options are:

- `fill` [string]: e.g., `blue`
- `fillOpacity` [number]: e.g., `0.3`
- `stroke` [string], e.g., `yellow`
- `strokeOpacity` [number], e.g., `0.6`
- `strokeWidth` [number], e.g., `1`
- `strokePos` [string, array] , e.g., `top` (`undefined` means top, left, right, bottom)
- `outline` [string], e.g., `cyan`
- `outlineOpacity` [number], e.g., `0.1337`
- `outlineWidth` [number], e.g., `1337`
- `outlinePos` [string, array], e.g., `['left', 'bottom']`

> Why is it necessary?

To be able to better customize overlays and make them equivalent to 1d annotations

Screen Shot 2019-10-02 at 8 50 23 PM

## Checklist

- [x] Updated CHANGELOG.md


app/scripts/OverlayTrack.js
```diff
@@ -11,7 +11,130 @@ class OverlayTrack extends PixiTrack {
...
+    isVertical = false
+  ) {
+    let finalXPos = xPos;
+    let finalYPos = yPos;
+    let finalWidth = width;
+    let finalHeight = height;
+
+    const outlineWidth = outline.width * 2 > width
+      ? width / 2 : outline.width;
+    const outlineHeight = outline.width * 2 > height
```
pkerpedjiev: Is this right? Should it be `outline.height * 2 > height`?

In either case, a comment above these lines describing what they're trying to accomplish would be helpful.

app/scripts/OverlayTrack.js
```diff
@@ -11,7 +11,130 @@ class OverlayTrack extends PixiTrack {
     this.drawnRects = {};
   }
 
-  drawHorizontalOverlay(graphics, position, extent, minWidth = 0) {
+  drawWithPositionedBorder(
```
pkerpedjiev: This doesn't have to be a member function of the class, does it? It has no references to `this`.

app/scripts/OverlayTrack.js
flekschas: Agreed, I'll add a comment. Apart from that this is correct because `outline.height` doesn't exist. The size of the stroke and outline is one dimensional so width == height.

app/scripts/OverlayTrack.js
flekschas: Good catch. Moved it outside as a context-free function.

flekschas:
Regarding #1: Good point! I'll update the docs.
Regarding #2: That's definitely something we should discuss in general. Right now I feel like most properties are named after SVG properties so switching the ones here to CSS equivalents would be odd. Having said that, maybe we should create a document where we go over all options to see whether SVG or CSS names are more appropriate. Regarding `borderLeft` etc. this would require more work because we can't specify all styles in one string at the moment and specifying 3 times the same with just because the position should be top, right, left, seem too verbose.

flekschas:
Update the docs :)

pkerpedjiev:
Looks good and thanks for the extra changes!

One last thing... would you mind just copying the options list from the PR description to the docs? I find the type annotations quite helpful and referred back to them to understand how the `strokePos` and `outlinePos` parameters should be filled out.

PR #797, opened 2019-11-01
Flekschas/fix event listener scrollable views

flekschas:
## Description

I added a second property to the global options object to be able to switch between scrolling and pan&zoom. I also fixed mouse interaction AFTER a view was scrolled.

Here is an example **with interactions** after scrolling:
Nov-01-2019 12-08-02

As discussed in Slack I refactored the states into `sizeMode` as follows.

_Note that the `bounded` property still exists for backward compatibility._

There are now 4 sizing modes:
  - ``default``: the height is given by the sum of the tracks' heights
  - ``bounded``: tells the HiGlass component to bind the height to the parent container by dynamically adjusting the height of center tracks.
  - ``scroll``
  - ``overflow``: the height of the views is given by the sum of the tracks' heights but when this height is larger than ``element`` the overflowing parts will be hidden. This mode is only needed when you want to dynamically switch between scrolling and pan+zooming. Say you scrolled halfway down and then want to temporarily pan&zoom a track at that position. If you would switch back to ``bounded`` the scrollTop position would be lost because ``bounded`` demands that your entire view is bound to the parent. Instead you want can switch to ``overflow`` to keep the current scrollTop position and enable pan&zooming.

> Why is it necessary?

Previously there were 2 issues when using scroll mode together with pan&zoom (separately but switching from one to the other)
1. The scroll position was lost after unsetting `scrollable` and blurring+refocussing the browser window
2. After scrolling, mouse interactions didn't work because the `mouseMoveHandler` didn't incorporate the scroll offset

Both issues are fixed now and tests are in place.

## Checklist

- [x] Updated CHANGELOG.md


CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+## v1.7.2
+
+- Add support for propper switching between scrolling and pan&zoom when `scrollable: true` by adding a new property called `scrolling`. When enabling `scrollable` HiGlass will adjust the style properties of it's root container to stretch to the full extent of its parent. When setting `scrolling: true` HiGlass will enable the appropriate CSS overflow rules. In summary, to activate scrolling set `scrollable: true` and `scrolling: true`. To temporarily switch back to pan&zoom keep `scrollable: true` but set `scrolling: false`. To fully reset HiGlass to the default pan&zoom mode set `scrollable: false` and `scrolling: false`.
```
pkerpedjiev: I think this is outdated given the new size mode option.

app/scripts/HiGlassComponent.js
```diff
@@ -359,6 +363,13 @@ class HiGlassComponent extends React.Component {
     }
   }
 
+  get sizeMode() {
+    // eslint-disable-next-line no-nested-ternary
+    return typeof this.props.options.sizeMode === 'undefined'
```
pkerpedjiev: this would probably be much clearer as an if statement.

flekschas:
Yes to both. :)

CHANGELOG.md
flekschas: Oh, good catch! I'll update it

app/scripts/HiGlassComponent.js
flekschas: Mhhh the if statement would look the same and it would lead to code duplication, which is the reason why I turned it into a function.

PR #805, opened 2019-11-20
BED12 annotations track support

alexpreynolds:
## Description

This adds a `HorizontalGeneBED12AnnotationsTrack` class and `gene-bed12-annotation` datatype for use with bigBed (BED12) sources.

## Checklist

- [x] Updated CHANGELOG.md


alexpreynolds:
I would like to ask for advice on unit testing. 

There is a `BedLikeTests.js` file that I could perhaps work from, except that I do not have a bigBed file hosted from an official higlass server (with its own uuid). 

I could use my development higlass-server instance, but if/when that server is taken down, any tests I build from it will fail.

Thanks for your help and advice!

pkerpedjiev:
Hey Alex, thanks for the PR! Using a production server to create the tests is a not so great habit that we've fallen into. In a better world, we would be mocking out the server return values for tile fetching.

With that in mind, there's two paths forward:

1. Add a bigBed dataset to higlass.io to use for testing.
2. Mock out the tile fetching in tileProxy to return the expected tile data.

In an ideal world #2 would be a wiser choice because it would not only make the tests independent but also set us on the right path for future tests. In the real world where your time is valuable and we've already set a bad example, #1 is acceptable.

alexpreynolds:
To help with testing, I'm going to investigate [fetch-mock](http://www.wheresrhys.co.uk/fetch-mock/) in `app/scripts/services/tile-proxy.js`, using mocked `fetch` calls to return expected tilesets for given parameters from the unit tests.

Obviously it would be great to have a working alternative approach that you are happy with. If this package works and is maintainable, great. If not, I might go back to option #1 (bother you folks about putting a bigBed file on the production server).

PR #816, opened 2019-12-31
Added valueScaleMin / valueScaleMax options ...

pkerpedjiev:
…and standardized formatting a little bit

## Description

As the title states.

> Why is it necessary?

These options should be documented.

Fixes #___

## Checklist

- [ ] Updated CHANGELOG.md

### DIFF
--- a/docs/track_types.rst
+++ b/docs/track_types.rst
@@ -231,7 +231,7 @@ Options
     }
   }
 
-- **valueScaleMin/valueScaleMax**: Absolute values limiting the value to y-position scale.
+- **valueScaleMin/valueScaleMax**: Absolute values limiting the the value scale, which is used to determine y-position (in 1D tracks) or color (heatmap) tracks.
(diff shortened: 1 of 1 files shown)


docs/track_types.rst
```diff
@@ -228,6 +231,7 @@ Options
     }
   }
 
+- **valueScaleMin/valueScaleMax**: Absolute values limiting the value to y-position scale.
```
flekschas: I am not sure I'd fully understand the description if I wasn't a core developer. How about:

Absolute values limiting the value scale, which is used to determine the y-position.

Just an idea :)

pkerpedjiev:
```suggestion
- **valueScaleMin/valueScaleMax**: Absolute values limiting the limits of the value scale, which is used to determine y-position (in 1D tracks) or color (heatmap) tracks.
```

PR #819, opened 2020-01-07
[WIP] added docs for docker networking config example

sergpolly:
## Description

added docker networking config example - trying to address following issues:
https://github.com/higlass/higlass-manage/issues/34
https://github.com/higlass/higlass-docker/issues/148
@pkerpedjiev  - I did it finally! sorry for the delay.

> why is this neccessary ?

what happened to us was, that the default docker's IP-pool `172.17.x.x` was interfering with the WiFi router - which has an IP from the same subrange.
So docker was modifying the `route` table like so:
```
172.17.0.0      0.0.0.0         255.255.0.0     U     0      0        0 docker0
172.18.1.0      0.0.0.0         255.255.255.0   U     0      0        0 br-blahblahblah
```
and then the machine running this docker would get unavailable via Wi-Fi ...
Pretty nasty issue, and a bit hard to figure out for someone unfamiliar with the networking.

This isn't very well documented on the docker side unfortunately, but the config-capabilities are there.
Here are all relevant links:
https://docs.docker.com/engine/reference/commandline/dockerd/
https://forums.docker.com/t/custom-ip-range-for-new-networks/60839
https://docs.docker.com/engine/reference/commandline/network_create/

PS
altering global docker config in `/etc/docker/daemon.json` seems like a great solution because it would restrict IP-addresses used by docker in general - useful when using `redis` and maybe some other docker-containers alongside higlass ...
The previous solution https://github.com/higlass/higlass-docker/issues/148 by adding `docker create network --subnet x.x.x.x` would not take care of all the cases

PPS
I have a feeling information like that belong more to something like `troubleshooting` section of some kind, but there is no such thing yet AFAIK

Fixes #___

added docs - provided useful example

## Checklist

- [ ] Updated CHANGELOG.md

### DIFF
--- a/docs/higlass_docker.rst
+++ b/docs/higlass_docker.rst
@@ -48,3 +48,20 @@ To use the admin interface for managing the available datasets, a superuser need
   docker exec -it higlass-container higlass-server/manage.py createsuperuser
 
 Once a username and password are created, the admin interface can be accessed at ``http://localhost:8989/admin``.
+
+Networking configuration
+************************
+
+When setting up higlass-docker locally or remotely, docker creates virtual netowrking interafces that utilize real IP-addresses and modify host's routing table. Default pools of IP-addresses used by docker may create local networking interferences (e.g. with Wi-Fi routers, VPN-servers, etc).
+One could avoid such interferences by configuring pool of IP-addresses that are utilized by docker ``/etc/docker/daemon.json``, e.g.:
+
+.. code-block:: json
+
+   {
+       "default-address-pools":
+        [
+          {"base":"172.40.0.1/16","size":24}
+        ]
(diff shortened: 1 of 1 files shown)


pkerpedjiev:
This is great! Thanks for submitting this. Is it still in progress or can it be merged?

sergpolly:
I think it can be merged - it's tiny docs update - no code changes.
My only concern is that `Docker` section of higlass docs is not the most appropriate one for this `networking config`-thing.
Sounds more like troubleshooting to me

pkerpedjiev:
I think it's actually fine. We don't have a troubleshooting section and a Google search should land whoever is looking for it right on that page.

PR #840, opened 2020-01-23
Ticks that mark the bounds of the view

pkerpedjiev:
## Description

Added the `tickPosition` and `tickFormatting` options to the `HorizontalChromosomeLabels` track. See the updated docs for a description of the available options and what they mean.

> Why is it necessary?

Paper figures often show the chromosome positions at each end of the figure to precisely define which region is being shown. The `tickPosition` gives the chromosome labels track the ability to do that.

The `tickFormatting` option lets us more easily legible locations such as 10.2M instead of 10,214,881. This is important to have when using `tickPosition == 'ends'` because those values tend to be large and overly precise

### Screenshot

New tick positioning on top:

image

Fixes #___

## Checklist

- [x] Updated CHANGELOG.md


app/scripts/HorizontalChromosomeLabels.js
```diff
@@ -61,6 +62,18 @@ class HorizontalChromosomeLabels extends PixiTrack {
...
+    // we only draw two ticks on the left and the right of the screen
+    this.gBoundTicks = new PIXI.Graphics();
+
+    this.leftBoundTick = new PIXI.Text('', this.pixiTextConfig);
+    this.rightBoundTick = new PIXI.Text('', this.pixiTextConfig);
+
+    this.gBoundTicks.addChild(this.leftBoundTick);
+    this.gBoundTicks.addChild(this.rightBoundTick);
+
+    this.pMain.addChild(this.gBoundTicks);
```
flekschas: I wonder if it makes sense to wrap this into a function called `initBoundTicks()` together with a function `removeBoundTicks()` and only create these graphics objects when needed.

It might seem like overkill but I started noticing that we're already have an enormous amount of graphics objects in our scenegraph and it's starting to slow down things when there are a lot of tracks. Since `this.gBoundTicks` is essentially unused unless someone sets `tickPosition == 'ends'` I think we shouldn't create them unless needed. Or maybe we can reuse the same graphics object for even and end ticks?

pkerpedjiev: Makes sense, I'll update the PR 👍

PR #842, opened 2020-01-24
Horizontal multivec row filtering

keller-mark:
## Description

This adds the HorizontalMultivec track option `selectRows` to specify indices of the rows that should be displayed.

> Why is it necessary?

This feature will be used by the cistrome wrapper component for filtering and reordering samples.

## Checklist

- [x] Updated CHANGELOG.md

### Questions

- Right now the value of this option can be dynamically updated via the js api by manipulating the viewconf object and calling `.setViewConfig()`. Would there be a more efficient way to do this?
    - I tried adding the following function to `HorizontalMultivecTrack.js`
    ```
    selectRows(newSelection) {
       this.options.selectRows = newSelection;
       this.rerender();
    }
    ```
    but when I call `api.getTrackObject(viewId, trackId).selectRows()` I get an error that `this.options` is undefined. Also, a function like this would cause the option value in the viewconfig to become out of sync.
- What would be the best way to test this? Are there any tests that you can point me to that do some kind of visual diffing?


pkerpedjiev:
> I tried adding the following function to HorizontalMultivecTrack.js

Tracks should never set their own options. You pointed out the reason why yourself: the viewconf would become out of sync and it would introduce a bunch of extra complexity to keep it consistent. So the best solution is to update the viewconf object and let the effects percolate to the tracks. Does that not work?

> What would be the best way to test this? Are there any tests that you can point me to that do some kind of visual diffing?

We don't do any visual diffing anywhere. You're welcome to try to introduce it. Or you could see if there's a way to get the dimensions of the texture generated for each sprite at `tile.sprite` and make it sure it shrinks if you set `selectRows` to a subset of the available rows.

test/HorizontalMultivecTests.js
```diff
@@ -105,6 +106,35 @@ describe('Horizontal heatmaps', () => {
...
+          hgc.instance(),
+          'UiHlCoxRQ-aITBDi5j8b_w',
+          'YafcbvKDQvWoWRT1WrygPA'
+        ); // uuid of horizontal-multivec
+        const trackTiles = track.visibleAndFetchedTiles();
+        expect(trackTiles.length).toEqual(2);
+        expect(trackTiles[0].canvas.width).toEqual(256);
+        expect(trackTiles[0].canvas.height).toEqual(10);
+        expect(trackTiles[1].canvas.width).toEqual(256);
+        expect(trackTiles[1].canvas.height).toEqual(10);
```
pkerpedjiev: Thanks for the tests! They'd be even more effective if you created two tracks, one with the selectRows option and one without and checked `canvas.height`. That would provide a control.

app/scripts/worker.js
```diff
@@ -106,47 +108,70 @@ export function workerSetPix(
...
-      rgbIdx = 255;
-
-      if (
-        // ignore the upper right portion of a tile because it's on the diagonal
-        // and its mirror will fill in that space
-        !(ignoreUpperRight && Math.floor(i / tileWidth) < i % tileWidth) &&
-        !(ignoreLowerLeft && Math.floor(i / tileWidth) > i % tileWidth) &&
-        // Ignore color if the value is invalid
-        !Number.isNaN(+d)
+    if (filteredSize !== size) {
```
pkerpedjiev: Would you mind adding some comments here explaining what you're doing? It's clear to me while reviewing the PR but it might be puzzling to somebody that comes along later without that context.

app/scripts/worker.js
```diff
@@ -81,7 +81,9 @@ export function workerSetPix(
   pseudocount,
   colorScale,
   ignoreUpperRight = false,
-  ignoreLowerLeft = false
+  ignoreLowerLeft = false,
+  shape = null,
+  selectedRows = null
```
pkerpedjiev: Sorry, a bit of "do as I say, not as I do here", but would you mind documenting the extra parameters you added?

app/scripts/services/tile-proxy.js
```diff
@@ -560,7 +561,9 @@ export const tileDataToPixData = (
     pseudocount,
     colorScale,
     ignoreUpperRight,
-    ignoreLowerLeft
+    ignoreLowerLeft,
+    tile.tileData.shape,
+    tile.selectedRows
```
pkerpedjiev: I'm not sure that `selectedRows` should be a property of the tile. It's an option like `ignoreUpperRight` and `ignoreLowerLeft` that applies to all tiles equally, right?

app/scripts/HorizontalMultivecTrack.js
```diff
@@ -265,4 +275,15 @@ export default class HorizontalMultivecTrack extends HeatmapTiledPixiTrack {
...
   }
+
+  /**
+   * Render / draw a tile.
+   * Also set the variable tile.selectedRows.
+   *
+   * @param {Object}  tile  Tile data to be rendered.
+   */
+  renderTile(tile) {
+    tile.selectedRows = this.options.selectRows;
```
pkerpedjiev: See comment about `tile.selectedRows` below.

app/scripts/services/tile-proxy.js
keller-mark: Yes that is correct

app/scripts/worker.js
keller-mark: Sure thing

test/HorizontalMultivecTests.js
keller-mark: Definitely

PR #849, opened 2020-02-06
Add axis label format 'automatic'

alexander-veit:
## Description

The axis label format option 'automatic' was added which corresponds to the current option 'scientific'. 'automatic' is the new default. 'scientific' has been changed to be strictly exponential notation.

> Why is it necessary?

Make axis format options more clear.

Note: This is a breaking change. Existing view configs have to be updated in order to use the same label format as before.

## Checklist

- [x] Updated CHANGELOG.md


alexander-veit:
The fact that this is a breaking change is of course not great. Not so sure anymore if it is worth it.

flekschas:
Is there a real need for this change besides making the axis format more explicit? I.e., have people complained about the current behavior?

If not I would not be in favor of not introducing yet another option.

On the other hand, if this is something people want, how about introducing a new _"strict"_ mode? This boolean option could be called `axisLabelFormattingStrict` and be set to `false` by default. That would allow us to introduce strict formatting without introducing a breaking change. It would also allow us to have multiple formats that can have strict and automatic behavior.

alexander-veit:
Yeah, I don't think anybody complained. This just would be the cleanest way to do it in my opinion. `axisLabelFormattingStrict` would be a way to avoid the breaking change but it adds complexity.

The whole thing is probably not important enough to justify a breaking change for now. Maybe it could be incorporated at a time when there are some other major changes to the view config and users have to review them anyway?

manzt:
> The whole thing is probably not important enough to justify a breaking change for now. Maybe it could be incorporated at a time when there are some other major changes to the view config and users have to review them anyway?

We're ramping up work for a [higlass v2](https://github.com/higlass/higlass/pulls?q=sort%3Aupdated-desc+is%3Apr+label%3Av2+), and this sounds like a feature that would make sense for that batch.

Otherwise, since this PR has been open for over a year, I'd be interested in closing.

PR #852, opened 2020-02-07
Zero value color option

### LINKED ISSUE #851: Option zeroValueColor for horizontal-multivec track
_Commit_: b13d154

#### Steps to reproduce
1. Visit http://higlass.io/app/?config=e-gJlHZPQ5S3GjCHA86ZMA

#### Observed behavior
- Zeros in the data are encoded as white heatmap cells. The user would like to customize this behavior.

#### Expected behavior
- Zeros in the data are encoded using the value of `track.options.zeroValueColor` if it has been set in the view config.

keller-mark:
## Description

This adds the option `zeroValueColor` for the horizontal-multivec track.

> Why is it necessary?

Users may want to customize the color mapping for zero values.

Fixes #851 

## Checklist

- [x] Updated CHANGELOG.md


docs/track_types.rst
```diff
@@ -502,6 +502,7 @@ Options
 
 - **colorbarPosition**: ['hidden', 'topLeft', 'topRight', 'bottomLeft', 'bottomRight'] - The position of the colorbar element.
 - **colorbarBackgroundColor**: The background color for the colorbar element.
+- **zeroValueColor**: The color to use for zero data values. By default, white or black depending on the theme.
```
pkerpedjiev: What was it before you added this change?

pkerpedjiev:
Actually, I just realized that this is more useful than I originally thought. If it's not too much trouble, do you think you could add this option to regular heatmaps as well?

Here's an example: http://higlass.io/l/?d=d43xOBm2RXmAlQ13Ag5nBQ

You should see two faint dots in that view. If you zoom in, they'll disappear because the color scale will be from 2 to 2 and be completely white. Having a zero color would make this example work as expected.

And it would be nice to also have this selectable in the track's config menu. That's really easy to add, you just have to add the option's name to the track info for each track it should apply to:

https://github.com/higlass/higlass/blob/develop/app/scripts/configs/tracks-info.js

And then define the available options in the menu. Should be as simple as copying one of these entries and giving it a name:

https://github.com/higlass/higlass/blob/6cfce1d0e24a0c43c40684e211a3a67e8feb59ce/app/scripts/configs/options-info.js#L248

keller-mark:
Yes I will look into that tomorrow. Thanks for pointing me to the options menu info.

And I am getting some kind of chrom-sizes error for the example:

Screen Shot 2020-02-24 at 8 23 11 PM

pkerpedjiev:
You can ignore that error. It's the genome position search box looking for chromosome sizes to use.

keller-mark:
Ah I see now, I did not realize I just had to zoom out in the example, and that that was the whole point. I added the option to the `heatmap` track type and options menu config in the latest commit

keller-mark:
> Looks good.
> 
>     * How do we currently render zero values?

Currently they are rendered based on the colorScale function

> 
>     * Is it possible to make zero values transparent?

I have added this functionality here https://github.com/higlass/higlass/pull/852/commits/a0cca45b5cbab1763c09a56c803db0fb4b236841#diff-96ac9aced876173486b7f83ac8d7ace2R5

> 
>     * The zero color is different from the NaN color? If so would you mind clarifying that in the docs?

If the value is NaN then no value is set in pixData https://github.com/higlass/higlass/blob/develop/app/scripts/worker.js#L154

app/scripts/services/tile-proxy.js
```diff
@@ -494,13 +494,13 @@ export const calculateTilesFromResolution = (
  * @param valueScaleType: Either 'log' or 'linear'
  * @param valueScaleDomain: The domain of the scale (the range is always [254,0])
  * @param colorScale: a 255 x 4 rgba array used as a color scale
- * @param synchronous: Render this tile synchronously or pass it on to the
+ * @param synchronous: Render this tile synchronously or pass it on to the threadpool.
```
pkerpedjiev:
```suggestion
 * @param synchronous: Render this tile synchronously or pass it on to the threadpool (which doesn't exist yet).
```

PR #856, opened 2020-02-11
Fixed the track displacement issue when saving view configs

### LINKED ISSUE #855: Saving view config in the editor displaces tracks
#### Steps to reproduce
1. Go to http://higlass.io/
2. zoom somewhere
3. Open view config editor
4. Click "save"

#### Observed behavior
- Tracks are moved (panned and zoomed) on save, even though the view config is the same.

#### Expected behavior
- Tracks should remain at the same position

viewconf-editor

alexander-veit:
## Description

This PR resets the zoomTransform in the TrackRenderer when a new viewConfig is loaded, e.g., via the view config editor.

> Why is it necessary?

When the zoomTransform is not reset, saving a modified or unmodified view config via the editor can lead to an unexpected displacement of the existing tracks (since the zoomTransform is applied again).

Fixes #855 

viewconf-editor-fixed

## Checklist

- [ ] Updated CHANGELOG.md

### DIFF
--- a/app/scripts/TrackRenderer.js
+++ b/app/scripts/TrackRenderer.js
@@ -381,6 +381,18 @@ class TrackRenderer extends React.Component {
   }
 
   componentDidUpdate(prevProps) {
+    // If the initial domain changed, a new view config
+    // probably has loaded. Reset the zoom in this case.
+    if (
+      this.props.initialXDomain[0] !== prevProps.initialXDomain[0] ||
+      this.props.initialXDomain[1] !== prevProps.initialXDomain[1] ||
+      this.props.initialYDomain[0] !== prevProps.initialYDomain[0] ||
+      this.props.initialYDomain[1] !== prevProps.initialYDomain[1]
+    ) {
+      this.removeZoom();
+      this.addZoom();
+    }
+
     if (prevProps.isRangeSelection !== this.props.isRangeSelection) {
       if (this.props.isRangeSelection) {
         this.removeZoom();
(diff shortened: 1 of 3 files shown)


app/scripts/TrackRenderer.js
```diff
@@ -381,6 +381,17 @@ class TrackRenderer extends React.Component {
...
   componentDidUpdate(prevProps) {
+    // If the initial domain changed, a new view config
+    // probably has loaded. Reset the element's zoomTransform in this case.
+    if (
+      this.props.initialXDomain[0] !== prevProps.initialXDomain[0] ||
+      this.props.initialXDomain[1] !== prevProps.initialXDomain[1] ||
+      this.props.initialYDomain[0] !== prevProps.initialYDomain[0] ||
+      this.props.initialYDomain[1] !== prevProps.initialYDomain[1]
+    ) {
+      this.element.__zoom = zoomIdentity;
```
flekschas: A short comment what `__zoom` is used for would be useful in case we look at the code in the future :)

app/scripts/TrackRenderer.js
```diff
@@ -574,6 +585,10 @@ class TrackRenderer extends React.Component {
     this.yDomainLimits = yDomainLimits;
     this.zoomLimits = zoomLimits;
 
+    // Reset the zoomTransform, if the initial domain changed
```
flekschas: The " if the initial domain changed" is slightly confusing here as there is no `if` clause. I would just say, reset the zoom transformation on the HiGlass side to avoid future pan&zoom jumps. Or something like that.

PR #857, opened 2020-02-13
Fix corrupted locks and svg export bugs

pkerpedjiev:
## Description

This PR fixes a couple small but important bugs:

- If the locks are corrupted and a lock group doesn't contain a view that it should, an error will be thrown and the component won't load
- Exporting a SVG of a heatmap with no loaded tiles leads to an error

> Why is it necessary?

Errors bad. Stability good.

Fixes #___

## Checklist

- [x] Updated CHANGELOG.md


app/scripts/HeatmapTiledPixiTrack.js
```diff
@@ -855,11 +855,17 @@ class HeatmapTiledPixiTrack extends TiledPixiTrack {
...
-        'style',
-        `fill: rgb(${color[0]}, ${color[1]}, ${color[2]})`
-      );
+      if (color) {
+        rectColor.setAttribute(
+          'style',
+          `fill: rgb(${color[0]}, ${color[1]}, ${color[2]})`
+        );
+      } else {
+        // color hasn't been assigned so make this transparent
```
pkerpedjiev:
```suggestion
        // when no tiles are loaded, color will be undefined and we don't want to crash
```

test/LockTests.js
```diff
@@ -0,0 +1,195 @@
...
+  let div = null;
+
+  beforeAll(done => {
+    [div, hgc] = mountHGComponent(div, hgc, viewconf, done, {
+      style: 'width:800px; height:400px; background-color: lightgreen',
+      bounded: true
+    });
+  });
+
+  it('should export as SVG', () => {
```
keller-mark: Is this testing locks or SVG export? It seems that the file name does not match the test contents.

app/scripts/HiGlassComponent.js
```diff
@@ -2856,6 +2856,36 @@ class HiGlassComponent extends React.Component {
...
+        // we need to have the starting location of the
+        // lock's namesake view
+        toRemove.push(viewUid);
+      }
+    }
+
+    for (const viewUid of toRemove) {
+      if (locks[viewUid]) {
+        console.warn(
+          `View ${viewUid} not present in lockGroup: ${locks}. Removing...`
```
keller-mark: I am seeing the console warning 
```
View aa not present in lockGroup: [object Object]. Removing...
```
Should the locks object be stringified?

test/LockTests.js
pkerpedjiev: Yes, that's a good point. It actually does test the locks by virtue of loading the viewconf and not crashing but that requires some insider knowledge to realize. I added a couple tests to explicitly verify that the corrupt lock was removed.

app/scripts/HiGlassComponent.js
pkerpedjiev: Yeah... or removed. Knowing the view id that's absent should be enough to diagnose.

pkerpedjiev: I removed the object from the print statement.

PR #861, opened 2020-02-22
API .on('createSVG') callback option

keller-mark:
## Description

This adds the option to set a handler function for manipulating exported SVGs before they are downloaded.

> Why is it necessary?

This will allow us to include the extra visualizations from our cistrome-higlass-wrapper component in the SVGs that are exported from HiGlass using the existing "Export views as..." dropdown menu buttons.

## Checklist

- [x] Updated CHANGELOG.md

### DIFF
--- a/test/APITests.js
+++ b/test/APITests.js
@@ -1,6 +1,6 @@
 /* eslint-env node, jasmine */
 import { globalPubSub } from 'pub-sub-es';
-import { select } from 'd3-selection';
+import { select, create } from 'd3-selection';
 
 import {
   some,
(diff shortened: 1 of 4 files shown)


test/APITests.js
```diff
@@ -838,4 +838,66 @@ describe('API Tests', () => {
...
+  describe('Export SVG API tests', () => {
+    it('listens to create SVG events', done => {
+      [div, api] = createElementAndApi(simple1And2dAnnotations, {
+        editable: false,
+        bounded: true
+      });
+
+      api.on('createSVG', svg => {
+        expect(svg.children.length).toEqual(2);
+        done();
```
pkerpedjiev: done called before return?

keller-mark: I agree that it looks strange, but if it is after the return it will never be called, correct?

pkerpedjiev:
Are the docs there or can I just not find them?

keller-mark:
It looked like the JS API docs are just derived from the jsdoc comments https://github.com/higlass/higlass/pull/861/files#diff-e80a6b2d4c494c27266c4c575997a52eR820 but was not entirely clear

pkerpedjiev:
Oh yeah, you're right :-) Would you mind adding a small example of how it would be used? I.e. does it operate on the SVG string or the DOM node?

keller-mark:
I have added an example here https://github.com/higlass/higlass/pull/861/files#diff-e80a6b2d4c494c27266c4c575997a52eR852

pkerpedjiev-zymergen:
Right, thanks!

keller-mark:
This is an example of the randomly failing tests. All I changed was a comment in my commit https://travis-ci.org/higlass/higlass/builds/655457524#L1114

pkerpedjiev:
Yeah, I see what you mean about the tests. Unfortunately, I don't have a good solution right now. Mocking out the requests is straightforward but really tedious. Maybe just restart the tests for now and at some point we can dive in and refactor the tests.

Although I do wonder why that's failing... it might be an indication that the higlass.io server is under heavy load, which would be whole other problem to solve.

keller-mark:
Ok that makes sense. For now I will restart the tests, that usually works.

keller-mark:
Can this be merged or should I wait for approval from all reviewers?

pkerpedjiev:
Feel free to merge, generally one approval is enough, unless you think the change is really risky.

keller-mark:
Ok great, thanks

PR #864, opened 2020-02-24
Support null fromViewUid in viewport-projection tracks

keller-mark:
## Description

This pull request modifies the `viewport-projection-*` tracks to add support for null `fromViewUid` values. Instead, coordinates for the new `initialXDomain` and `initialYDomain` track properties can be provided to define the interval. When the d3-brush selection rect is moved, the values of `initialXDomain` and `initialYDomain`  in the viewConfig are updated, just like they are for the `initialXDomain` and `initialYDomain` values in view objects, and trigger the `.on('viewConfig')` event.

> Why is it necessary?

We are using the `viewport-projection-horizontal` in the cistrome wrapper component to enable selection of genomic intervals. However, we do not need to display a second view from which the domain for the tracks are linked.

## Checklist

- [x] Updated CHANGELOG.md

## Screenshot

Screen Shot 2020-02-24 at 5 50 49 PM


app/scripts/ViewportTrackerVertical.js
```diff
@@ -146,6 +145,21 @@ class ViewportTrackerVertical extends SVGTrack {
...
     this.draw();
   }
+
+  setDimensions(newDimensions) {
+    super.setDimensions(newDimensions);
+
+    const xRange = this._xScale.range();
+    const yRange = this._yScale.range();
+    const yDiff = yRange[1] - yRange[0];
+    this.brush.extent([
```
pkerpedjiev: What is this doing?

app/scripts/HiGlassComponent.js
```diff
@@ -2804,6 +2804,52 @@ class HiGlassComponent extends React.Component {
       track.removeViewportChanged = trackId =>
         this.removeScalesChangedListener(fromView, trackId);
       track.setDomainsCallback = (xDomain, yDomain) => {
+        if (!fromView) {
+          // If there is no `fromView`, then there must be an `initialXDomain` instead.
+          // Update the viewconfig to reflect the new `initialXDomain` array
+          // on the `viewport-projection-horizontal` track.
+          const { viewConfig } = this.state;
+          const newViewConfig = JSON.parse(JSON.stringify(viewConfig));
```
pkerpedjiev: Does this get called on every zoom event? If so, all this stringifying and parsing could have a significant effect on performance.

keller-mark: Yes you are correct, this was unnecessary

app/scripts/ViewportTrackerVertical.js
keller-mark: In `ViewportTracker2D`, the brushing extent was being set to `Number.MAX_VALUE` 
https://github.com/higlass/higlass/pull/864/files#diff-d04084b0801ca6d78773229b4e54ecffL30

which was causing issues when the rectangle was rendered, since the HTML attribute values (width, height, etc.) were on the order of 10^300.

On the other hand, in `ViewportTrackerVertical` and `ViewportTrackerHorizontal`, the extent was not being set at all, and the default extent (https://github.com/d3/d3-brush/blob/master/src/brush.js#L116) was causing the selection rectangles to be either too short or too skinny (I think because the default extent is based on the container dimensions, and the container was smaller before the viewconfig was fully loaded).

This change (for example in `ViewportTrackerVertical`) uses the scale range to set the extent. It sets the extent to the total y range (total height of the view), plus extra (total height / 2) on the top and bottom, to allow the brushing selection to move out of the view on the top or bottom, since the current behavior in higlass allows the brush to be moved outside the view, and I wanted to keep that behavior.

keller-mark:
I have the following use case for `projectionMarginTop`, and added the others for symmetry.
When using the `viewport-projection-horizontal`, I want to be able to put the track in the `whole` track position, and then prevent the brushing rect from covering a top `horizontal-chromosome-labels` track. If the `horizontal-chromosome-labels` track height is 30, I can set `projectionMarginTop` to 30 to do this.

pkerpedjiev:
> I have the following use case for projectionMarginTop, and added the others for symmetry.
When using the viewport-projection-horizontal, I want to be able to put the track in the whole track position, and then prevent the brushing rect from covering a top horizontal-chromosome-labels track. If the horizontal-chromosome-labels track height is 30, I can set projectionMarginTop to 30 to do this.

Hmmm... I'm not sure I get this entirely. If you add a margin to a viewport projection, then it's not showing the true viewport?

Also, which track are you putting in the 'whole' position?

keller-mark:
If you look at the screenshot: the red `viewport-projection-horizontal` track is in the "whole" position. But notice how it is not covering the topmost chromosome labels. That is because it has projectionMarginTop = 30. 

Also note that:
- `viewport-projection-horizontal` does not use projectionMarginLeft or projectionMarginRight
- `viewport-projection-vertical` does not use projectionMarginTop or projectionMarginBottom
- `viewport-projection-center` does not use any of the margin options

pkerpedjiev:
Does it have a bottom margin as well? Since it doesn't reach the bottom?

I guess I'm just concerned that this is adding an option that fixes a symptom but leaves the underlying problem unsolved.

The problem, as I understand it, is that you want the viewport projection to cover only some of the tracks. The solution implements that if you only want to leave the top track uncovered but fails if you'd like to leave middle tracks uncovered. Maybe take a look at how overlay tracks are implemented (http://docs.higlass.io/view_config.html#overlay-tracks) and see if a similar approach might not work here.

If my diagnosis is correct, then I would recommend breaking this PR up into two parts:

1. Support null in the `fromViewUid` - this we can merge right away
2. Adding support for selective track covering - this we can continue discussing

keller-mark:
Ok that is a good idea. I will remove the projectionMargin stuff from this PR

app/schema.json
```diff
@@ -406,10 +410,57 @@
...
+          "projectionYDomain": {
+            "$ref": "#/definitions/viewProperties/domain"
+          },
+          "local": { "type": "boolean" },
+          "maxWidth": { "type": "number" },
+          "name": { "type": "string" },
+          "options": { "type": "object" },
+          "orientation": { "type": "string" },
+          "position": { "type": "string" },
+          "project": { "type": "null" },
```
pkerpedjiev: This should not be in the viewconf.

app/schema.json
```diff
@@ -406,10 +410,57 @@
...
+          "project_name": { "type": "string" },
```
pkerpedjiev: This should also not be in the viewconf.

app/schema.json
```diff
@@ -406,10 +410,57 @@
...
+            "$ref": "#/definitions/viewProperties/domain"
+          },
+          "projectionYDomain": {
+            "$ref": "#/definitions/viewProperties/domain"
+          },
+          "local": { "type": "boolean" },
+          "maxWidth": { "type": "number" },
+          "name": { "type": "string" },
+          "options": { "type": "object" },
+          "orientation": { "type": "string" },
```
pkerpedjiev: This should not be in the viewconf.

app/schema.json
```diff
@@ -406,10 +410,57 @@
...
+          "fromViewUid": { "type": "null" },
+          "header": { "type": "string" },
+          "height": { "type": "number" },
+          "projectionXDomain": {
+            "$ref": "#/definitions/viewProperties/domain"
+          },
+          "projectionYDomain": {
+            "$ref": "#/definitions/viewProperties/domain"
+          },
+          "local": { "type": "boolean" },
```
pkerpedjiev: This should not be in the viewconf.

app/schema.json
```diff
@@ -406,10 +410,57 @@
...
+          "maxWidth": { "type": "number" },
```
pkerpedjiev: Not necessary for viewport projection tracks.

app/schema.json
```diff
@@ -406,10 +410,57 @@
...
+            "enum": [
+              "viewport-projection-horizontal",
+              "viewport-projection-vertical",
+              "viewport-projection-center"
+            ]
+          },
+          "chromInfoPath": { "type": "string" },
+          "created": { "type": "string" },
+          "defaultOptions": { "type": "object", "additionalProperties": false },
+          "description": { "type": "string" },
```
pkerpedjiev: Shouldn't be in the viewconf for viewport-projection-tracks.

app/schema.json
```diff
@@ -406,10 +410,57 @@
...
+          "fromViewUid": { "type": "null" },
+          "header": { "type": "string" },
```
pkerpedjiev: Shouldn't be in the viewconf for viewport-projection-tracks.

app/schema.json
```diff
@@ -406,10 +410,57 @@
...
+          },
+          "local": { "type": "boolean" },
+          "maxWidth": { "type": "number" },
+          "name": { "type": "string" },
+          "options": { "type": "object" },
+          "orientation": { "type": "string" },
+          "position": { "type": "string" },
+          "project": { "type": "null" },
+          "project_name": { "type": "string" },
+          "thumbnail": {
```
pkerpedjiev: Shouldn't be in the viewconf for viewport-projection-tracks.

app/scripts/HiGlassComponent.js
```diff
@@ -2911,6 +2911,52 @@ class HiGlassComponent extends React.Component {
...
+                        );
+                    }
+                  }
+                }
+              }
+            }
+          }
+
+          const newViews = this.processViewConfig(newViewConfig);
+          this.setState({ views: newViews, viewConfig: newViewConfig });
```
pkerpedjiev: I'm still not convinced that this is the right way to go about this. The viewport track's projection*Scales can change very rapidly so they shouldn't be part of the state. Why not store them in a separate variable e.g. this.viewportScales[viewportTrack.uid] and only add them to the viewconf on export? 

See `this.xScales` in `HiGlassComponent` for an example of how we accomplish this for locked scales in tracks.

Here's the key part where that information is added to viewconf on export: https://github.com/higlass/higlass/blob/90d2b29d09a9b1e8e6d79b205a0bb42c0803d802/app/scripts/HiGlassComponent.js#L3112

app/scripts/ViewportTrackerVertical.js
pkerpedjiev: 👍

app/scripts/HiGlassComponent.js
keller-mark: That does seem like a better approach. In 2a7d3f1 I have added `this.projectionXDomains` and `this.projectionYDomains` to store the changing scale domain information. 

I did also add a call to `this.triggerViewChangeDb` because I do still want the update to trigger the `.on('viewConfig')` callback

app/schema.json
```diff
@@ -406,10 +410,44 @@
...
+          "type": {
+            "enum": [
+              "viewport-projection-horizontal",
+              "viewport-projection-vertical",
+              "viewport-projection-center"
+            ]
+          },
+          "chromInfoPath": { "type": "string" },
+          "created": { "type": "string" },
+          "defaultOptions": { "type": "object", "additionalProperties": false },
```
pkerpedjiev: Sorry, I didn't notice this last time but it shouldn't be in here.

app/schema.json
```diff
@@ -406,10 +410,44 @@
...
+        "properties": {
+          "uid": { "type": "string" },
+          "type": {
+            "enum": [
+              "viewport-projection-horizontal",
+              "viewport-projection-vertical",
+              "viewport-projection-center"
+            ]
+          },
+          "chromInfoPath": { "type": "string" },
```
pkerpedjiev: This shouldn't be here.

app/schema.json
```diff
@@ -406,10 +410,44 @@
...
+          "created": { "type": "string" },
```
pkerpedjiev: This shouldn't be here.

app/schema.json
```diff
@@ -406,10 +410,44 @@
...
+          "defaultOptions": { "type": "object", "additionalProperties": false },
+          "fromViewUid": { "type": "null" },
+          "height": { "type": "number" },
+          "projectionXDomain": {
+            "$ref": "#/definitions/viewProperties/domain"
+          },
+          "projectionYDomain": {
+            "$ref": "#/definitions/viewProperties/domain"
+          },
+          "name": { "type": "string" },
```
pkerpedjiev: This shouldn't be here.

app/schema.json
```diff
@@ -406,10 +410,44 @@
...
+          "options": { "type": "object" },
+          "position": { "type": "string" },
```
pkerpedjiev: I don't think that this should be here either.

app/schema.json
```diff
@@ -406,10 +410,44 @@
...
+              "viewport-projection-horizontal",
+              "viewport-projection-vertical",
+              "viewport-projection-center"
+            ]
+          },
+          "chromInfoPath": { "type": "string" },
+          "created": { "type": "string" },
+          "defaultOptions": { "type": "object", "additionalProperties": false },
+          "fromViewUid": { "type": "null" },
+          "height": { "type": "number" },
```
pkerpedjiev: I don't think that this should be here.

app/scripts/HiGlassComponent.js
```diff
@@ -3088,6 +3116,23 @@ class HiGlassComponent extends React.Component {
           }
         }
 
+        if (
```
pkerpedjiev: You have a nice comment in the block above stating that it's executed if there's no `fromView`. Maybe another comment here would be helpful to refresh the readers memory that this block is entered if `fromView` is present?

PR #869, opened 2020-03-04
Major update of the Gene Annotations Track

pkerpedjiev:
## Description

- Thicker gene annotations by default
- Arrowhead at the tip
- Rendering of "filler" annotations
- Separation of different rendering methods
- Mouseover indicator
- Click handler

> Why is it necessary?

- Stylize 😎 
- Filler annotations are really useful for showing regions that have hidden annotations

Fixes #___

## Screengrab

**Before**

image

**After**

image

## Checklist

- [x] Updated CHANGELOG.md


test/HiGlassComponentTest.js
```diff
@@ -533,53 +533,6 @@ describe('Simple HiGlassComponent', () => {
     });
   });
 
-  describe('Gene Annotations Display', () => {
```
pkerpedjiev: Moved into GeneAnnotationTrackTests.js above.

karma.conf.js
```diff
@@ -27,49 +27,49 @@ module.exports = config => {
...
-      'test/ChromosomeLabelsTests.js',
-      'test/OverlayTrackTests.js',
-      'test/PngExportTest.js',
-      'test/ViewConfigEditorTests.js',
-      'test/ViewManipulationTests.js',
-      'test/ViewportProjectionTests.js',
-      'test/ZoomTests.js',
-      'test/GenbankFetcherTests.js',
-      'test/SchemaTests.js',
+      'test/AxisTests.js',
```
pkerpedjiev: Reordered these so that they are alphabetical.

app/scripts/utils/get-track-conf-from-hgc.js
```diff
@@ -0,0 +1,36 @@
+/**
```
pkerpedjiev: Added this utility method for testing.

app/styles/HiGlass.module.scss
```diff
@@ -50,15 +50,14 @@
...
-    margin: 17px 0 0 9px;
-    padding: 0 0.25rem;
-    font-size: 0.8em;
-    pointer-events: none;
-    background: white;
-    border-radius: 0.25rem;
-    box-shadow: 0 0 0 1px rgba(0, 0, 0, 0.1),
-                0 0 3px 0 rgba(0, 0, 0, 0.075),
-                0 0 7px 0 rgba(0, 0, 0, 0.05);
+  position: fixed;
```
pkerpedjiev: Automatic styling updates.

app/styles/PlotTypeChooser.css
```diff
@@ -1,14 +1,14 @@
 .plot-type-selected {
```
pkerpedjiev: More code style updates. No functional changes.

test/GenomePositionSearchBoxTest.js
```diff
@@ -298,5 +311,18 @@ describe('Genome position search box tests', () => {
     it("checks that the div hasn't grown too much", () => {
       expect(div.clientHeight).toBeLessThan(500);
     });
+
+    it('Cleans up', () => {
```
pkerpedjiev: This test was missing a cleanup method but I only just noticed after reordering the tests in karma.conf.js.a

test/GenomePositionSearchBoxTest.js
```diff
@@ -74,6 +74,19 @@ describe('Genome position search box tests', () => {
         });
       });
     });
+
+    it('Cleans up', () => {
```
pkerpedjiev: This test was missing a cleanup method but I only just noticed after reordering the tests in karma.conf.js.a

keller-mark:
Did you intend for the graphics (triangles) on the vertical gene annotations track to be larger than before?

For example these are using the same view config (the left is localhost running your branch, the right is higlass.io)

Screen Shot 2020-03-06 at 9 25 01 AM

keller-mark:
Also, why is there an option called `geneAnnoHeight` for vertical-gene-annotations and `geneAnnotationHeight` for horizontal-gene-annotations? It seems like these options should have the same name if they do the same thing.

https://github.com/higlass/higlass/pull/869/files/2d387de1972ddd30d9ac0c399ed5c5d9cf54c288#diff-e0ca334687e5bd05fb4845c71c225fb7R1260

test/GeneAnnotationsTrackTests.js
```diff
@@ -0,0 +1,136 @@
...
+        clientY: 117,
+        screenX: 278,
+        screenY: 231
+      };
+
+      canvasElem.dispatchEvent(createPointerEvent('pointerdown', loc));
+      canvasElem.dispatchEvent(createPointerEvent('pointerup', loc));
+
+      expect(clicked).to.eql(1);
+      // add your tests here
```
keller-mark: Is this comment necessary?

test/GeneAnnotationsTrackTests.js
```diff
@@ -0,0 +1,136 @@
...
+  it('exports to SVG', () => {
+    const svgStr = hgc.createSVGString();
+
+    expect(svgStr.indexOf('path')).to.be.above(0);
+    expect(svgStr.indexOf('text')).to.be.above(0);
+  });
+
+  afterAll(() => {
+    // hgc.instance().handleTrackOptionsChanged('v', 'heatmap0', newOptions0);
+    // removeHGComponent(div);
```
keller-mark: Can these lines be removed?

test/GeneAnnotationsTrackTests.js
```diff
@@ -0,0 +1,136 @@
...
+    waitForTilesLoaded(hgc, () => {
+      // make sure the gene is red
+      expect(trackObj.allTexts[0].text.style.fill).to.eql('#ff0000');
+
+      const trackConf = getTrackConfFromHGC(hgc, viewUid, trackUid);
+      const options = trackConf.options;
+
+      // set minus strand genes to black
+      options.minusStrandColor = 'black';
+      // const trackConfig = getTrackByUid;
```
keller-mark: Can this line be removed?

docs/examples/examples.html
```diff
@@ -1,91 +1,1576 @@
...
+          <a href="apis/svg.html?/viewconfs/1d-annotations.json"
+            >1d-annotations.json</a
+          >
+        </td>
+        <td></td>
+        <td></td>
+        <td></td>
+        <td>X</td>
+        <td></td>
+        <td></td>
```
keller-mark: I understand that this is just a code style change and not a big deal but maybe a separate issue can be added to change this table so that it is generated in javascript based on some more manageable data structure?

app/scripts/utils/get-track-conf-from-hgc.js
```diff
@@ -0,0 +1,36 @@
...
+
+  Object.keys(tracks)
+    .map(trackType => tracks[trackType])
+    .filter(tracksByPos => tracksByPos.filter)
+    .reduce((a, b) => a.concat(b), [])
+    .forEach(checkTrack);
+
+  if (found) return found;
+
+  return null;
```
keller-mark: Super minor, but if found is initially set to null, can you just return `found` and remove line 31?

app/scripts/configs/tracks-info.js
```diff
@@ -1157,7 +1157,7 @@ export const TRACKS_INFO = [
     type: 'horizontal-gene-annotations',
     datatype: ['gene-annotation'],
     local: false,
-    defaultHeight: 55,
+    defaultHeight: 90,
```
keller-mark: Should `defaultWidth` for `vertical-gene-annotations` reflect this change?

app/scripts/HorizontalGeneAnnotationsTrack.js
```diff
@@ -187,163 +609,106 @@ class HorizontalGeneAnnotationsTrack extends HorizontalTiled1DPixiTrack {
...
-          fontFamily: FONT_FAMILY,
-          fill: fill[geneInfo[5]]
-        };
+    renderMask(this, tile);
 
-        if (!(geneId in tile.textWidths)) {
-          text.updateTransform();
-          const textWidth = text.getBounds().width;
-          const textHeight = text.getBounds().height;
+    stretchRects(this);
```
keller-mark: More of a code style preference, but it seems bad to just have functions where you pass `this` around.

The individual functions do not seem to be using many things from their `track` parameters, it looks like in total in all the functions, the only instance fields that are used are `track._xScale`, `track.renderTile`, `track.tilesetInfo`, `track.fetchedTiles`, `track.dimensions`.

Maybe one place where it is required is `geneClickFunc` where the track needs to be passed to the function

keller-mark:
Also: I see that you removed the following track properties from many of the test view configs, I am assuming to simplify them because they are unnecessary, which is great.
- name
- created
- project
- project_name
- description
- transforms
- resolutions
- position
- orientation
- local
- header

Possibly outside the scope of this PR, but I think that `resolutions` and `aggregationModes` should also be added as deleted properties here 
https://github.com/higlass/higlass/blob/develop/app/scripts/HiGlassComponent.js#L3155

since they show up in exported view configs (from `api.getViewConfig()` or `api.on('viewConfig')` or the editor) even though higlass schema validation always complains that they are "additionalProperties"

test/HiGlassComponentTest.js
keller-mark: Looks good

karma.conf.js
keller-mark: Nice

app/scripts/configs/tracks-info.js
pkerpedjiev: Yes it should indeed. Nice catch!

app/scripts/utils/get-track-conf-from-hgc.js
pkerpedjiev: Sure can!

test/GeneAnnotationsTrackTests.js
pkerpedjiev: Yup.

test/GeneAnnotationsTrackTests.js
pkerpedjiev: Yup.

pkerpedjiev:
Thanks for the comments!

> Possibly outside the scope of this PR, but I think that resolutions and aggregationModes should also be added as deleted properties here

Agreed, I added them into this PR. If you notice that happening again the future, please add a test to make sure we're catching them.

docs/examples/examples.html
pkerpedjiev: I think that might actually already exist. @mccalluc do you happen to remember?

app/scripts/HorizontalGeneAnnotationsTrack.js
pkerpedjiev: In the general case, I agree with you but I feel like in this particular instance there is some logical coherence and elegance to saying `stretchRects(track)` rather than `stretchRects(xScale, renderTile, fetchedTiles)`.

If you don't strongly object, I'll keep it like this and if it becomes troublesome we can refactor in the future.

test/GeneAnnotationsTrackTests.js
pkerpedjiev: Absolutely not.

pkerpedjiev:
> Also, why is there an option called geneAnnoHeight for vertical-gene-annotations and geneAnnotationHeight for horizontal-gene-annotations? It seems like these options should have the same name if they do the same thing.

Yup, good catch! I changed the option in verticalGeneAnnotations.

app/scripts/HorizontalGeneAnnotationsTrack.js
keller-mark: No strong objection, that sounds good to me.

docs/examples/examples.html
mccalluc: This is generated in https://github.com/higlass/higlass/blob/develop/docs/examples/regenerate.py ... I'm not sure it'd be worth while to just change it to a JSON representation: The real issue that it's built artifact which is checked in, and JSON would be the same in that regard... if we wanted to address this, perhaps reference regenerate.py from https://github.com/higlass/higlass/blob/develop/docs/build.sh?

keller-mark: Ok, yes probably not worth while. I will mark as resolved.

keller-mark:
Great

PR #871, opened 2020-03-06
Local tile fetcher

pkerpedjiev:
## Description

- This lets tracks use a local tile fetcher and specify raw tile values.
- Updates the the schema. Most of the changed files are just updates to viewconfs removing unnecessary fields

> Why is it necessary?

- Can make testing less reliant on having a server
- Can be used by applications that want to display a track containing only a few values that can be stored directly in the viewconf.

Fixes #___

## Checklist

  - Allowed additionalProperties for data
- [x] Updated CHANGELOG.md


app/schema.json
```diff
@@ -36,6 +36,16 @@
     "chromInfoPath": { "type": "string" }
   },
   "definitions": {
+    "data": {
+      "type": "object",
+      "properties": {
+        "type": { "type": "string" },
+        "tilesetInfo": { "type": "object" },
+        "children": { "type": "array" },
```
keller-mark: Just curious what this children field is used for. I see that it was part of the existing data object properties.

pkerpedjiev: It can be used to create tracks that display the ratio of two datasets. I just added a section to the docs explaining this: https://github.com/higlass/higlass/pull/871/files#diff-74d725479bc068778bbd4ad8e310a67dR352-R372

pkerpedjiev:
>  Just curious, how did you adjust all the test view confs to the updated schema?

Just went through and edited them. There was a period of time when we were exporting more fields than necessary. The change I made to the schema removed those fields and then I went though the test viewconfs and manually removed them.

app/schema.json
keller-mark: Ok thanks!

alexander-veit:
Ok. Thanks. I thought there is maybe an automated way that I was not aware of.

PR #874, opened 2020-03-10
Improvements to make the BedLikeTrack more useable

pkerpedjiev:
## Description

- Used `stretchRects` to ensure that there are no rendering artifacts when zooming in too far (test  ✔)
- Added `scaled` as a potential value to the `annotationHeight` option. (the height of the annotations is scaled so that they fit within the bounds of the track without overlaps) (test  ✔)

image

- Added the `maxAnnotationHeight` option to limit the size of annotations when using the `scaled` option for `annotationHeight`. (test  ✔)
- Added the `fontSize` option for bed-like tracks (test  ✔)
- Made `colorAnnotations` independent of `valueScale` (test  ✔)

> Why is it necessary?

Fixes #___

## Checklist

- [x] Updated CHANGELOG.md


test/BedLikeTests.js
```diff
@@ -105,6 +105,113 @@ describe('Simple HiGlassComponent', () => {
       done();
     });
 
+    it('Checks to make sure that scaled height changes the height of drawn rects', () => {
+      // const currentHeight =
```
keller-mark:
```suggestion
```

app/scripts/configs/tracks-info.js
```diff
@@ -1812,6 +1815,7 @@ export const TRACKS_INFO = [
       labelTopMargin: 0,
       labelBottomMargin: 0,
       minHeight: 20,
+      maxAnnotationHeight: 'none',
```
keller-mark: In some of my recent pull requests I have added track options, with `null` default values. Is using a "none" string preferred?

app/scripts/TiledPlot.js
```diff
@@ -484,14 +486,14 @@ class TiledPlot extends React.Component {
...
     // to speedup the subsequent filtering
-    const trackUids = allTracksWithTilesetInfos.reduce((a, b) => {
-      a[b.id] = true;
-      return a;
-    }, {});
-
-    // Only could tracks that are suppose to get a tileset
-    const loadedTilesetInfos = Object.keys(this.tracksByUidInit).filter(
-      trackUid => trackUids[trackUid]
+    // const trackUids = allTracksWithTilesetInfos.reduce((a, b) => {
```
keller-mark: Is there a reason to keep this commented rather than remove?

keller-mark:
Can you point me to a view config to visually test this? I tried the one in the test `test/BedLikeTests.js` file but did not see any rects (and not sure if I was supposed see any there)

app/scripts/configs/tracks-info.js
pkerpedjiev: No, I think you're right and null is actually better. Saves an extra if clause. The only time an explicit 'none' or something like that might be better is where 0 is a valid value.

app/scripts/TiledPlot.js
pkerpedjiev: Nope.

pkerpedjiev:
> Can you point me to a view config to visually test this? I tried the one in the test test/BedLikeTests.js file but did not see any rects (and not sure if I was supposed see any there)

Yeah, that viewconf has locally specified tiles that only appear at some zoom level. If you zoom in (or out, depending on your screen size), you should see them pop up.

Otherwise you can use the viewconf here: http://higlass.io/l/?d=cVG3jIhdSXy0XcKnHhkKow

keller-mark:
Tested locally and looks good.

pkerpedjiev:
Thanks!

PR #876, opened 2020-03-11
Publish wheel events

keller-mark:
## Description

This pull request adds support for a `.on('wheel')` callback to enable the wheel event to be published. It also adds an API function `.setWheelCallbackKeydownCondition(keyCode)`, which if used, ensures that the wheel callback is only called when the specified keyboard key has been pressed down.

> Why is it necessary?

This will enable us to catch the wheel event in the cistrome wrapper to support "zooming" along the y axis, which we will try to implement by listening for the wheel event and updating the `selectRows` array. (Though we are still not sure how smooth this approach will be)

Fixes #___

## Checklist

- [x] Updated CHANGELOG.md


keller-mark:
Your solution is much nicer, I have adopted that, thanks.

app/scripts/HiGlassComponent.js
```diff
@@ -820,12 +825,40 @@ class HiGlassComponent extends React.Component {
...
+
   keyDownHandler(event) {
     if (this.props.options.rangeSelectionOnAlt && event.key === 'Alt') {
       this.setState({
         mouseTool: MOUSE_TOOL_SELECT
       });
     }
+
+    if (this.isKeyDownWheelConditionSatisfied(event.key)) {
+      this.shouldPublishWheelEvent = true;
```
pkerpedjiev: What if you register two wheel event handlers? Won't this cause cross-talk?

keller-mark:
I am just going to remove the whole keydown condition thing. We should be able to listen for those events from outside higlass and then set / remove the wheel handler accordingly.

keller-mark:
@pkerpedjiev I removed the keydown parameter and now I just check if the number of wheel callbacks is greater than zero, where I increment / decrement with each call to .on / .off

https://github.com/higlass/higlass/pull/876/files#diff-e80a6b2d4c494c27266c4c575997a52eR891

Is that alright or is there a better approach?

app/scripts/HiGlassComponent.js
```diff
@@ -4288,6 +4291,52 @@ class HiGlassComponent extends React.Component {
...
+            : relTrackPos[0],
+        relTrackY:
+          hoveredTrack && hoveredTrack.flipText
+            ? relTrackPos[0]
+            : relTrackPos[1],
+        track: hoveredTrack,
+        origEvt: nativeEvent,
+        sourceUid: this.uid,
+        hoveredTracks,
+        noHoveredTracks: hoveredTracks.length === 0
```
pkerpedjiev: Why is this necessary? Can't the use just do the length check themselves?

app/scripts/api.js
```diff
@@ -682,6 +683,11 @@ const createApi = function api(context, pubSub) {
             apiPubSub.unsubscribe('mouseMoveZoom', callback);
             break;
 
+          case 'wheel':
+            self.numWheelCallbacks -= 1;
```
pkerpedjiev: I wonder if there's a way to just count the number of listeners in apiPubSub instead of maintaining another variable to keep track of that.

app/scripts/HiGlassComponent.js
keller-mark: It is not necessary, but it makes this event consistent with the existing mouse move event https://github.com/higlass/higlass/blob/develop/app/scripts/HiGlassComponent.js#L4044

pkerpedjiev: Interesting. OK, I'll hold my nose and let it slide :-)

app/scripts/api.js
keller-mark: Right now only the `publish` method is actually available in HiGlassComponent https://github.com/higlass/higlass/blob/develop/app/scripts/api.js#L32, so I will work on trying to make the listeners or number of listeners available as well

pkerpedjiev: OK, it's not essential but I think it would reduce a little bit of complexity.

PR #882, opened 2020-03-25
Enable zoom limits

### LINKED ISSUE #873: Zooming in past units of bases
_Commit_: 

- Current public build of higlass.io
- v1.8.5 higlass + v0.2.2 higlass-multivec (test development server)

#### Steps to reproduce

To reproduce, I pan and zoom into viewport until sub-base units are visible.

#### Observed behavior

We have found that multivec tracks on a test server (v1.8.5 higlass + v0.2.2 higlass-multivec) flicker and disappear when the user is zoomed in at fractional-base resolution, as well as the data in the Rao et al. Hi-C dataset.

Behavior on higlass.io:

Screen Recording 2020-03-06 at 3 44 42 PM

On test server:

Screen Recording 2020-03-06 at 4 02 27 PM

#### Expected behavior

Wouter and I chatted about this and, at the lowest level, genomic datasets are typically represented in units of bases. We couldn't think of datasets that would be expressed in smaller units. 

In the general case, it would be useful to lock zooming in past the point of a single base (or perhaps a specified resolution, for data that are coarser). This would also resolve the visual artifacts.

alexander-veit:
## Description

This PR adds the option to specify zoom limits in a view. When the `zoomLimits` field is present in the viewconf, zooming is only possible within these limits. E.g. if `"zoomLimits": [200, 3000000000]`, then highest possible zoom will always show 200bp (with respect to X) 

> Why is it necessary?

Users might want to avoid zooming in (or out) too far, e.g. zooming in past units of bases, in order to avoid visual artifacts.

Fixes #873 

## Checklist

- [ ] Updated CHANGELOG.md


pkerpedjiev:
Is it possible to specify just one limit? Like if I want to limit zooming in beyond a certain point but don't want to limit zooming out?

alexander-veit:
You mean, if we have `"zoomLimits": [200]` it should be treated as `"zoomLimits": [200, Infinity]`? I can easily add that.

alexpreynolds:
If it is possible to limit zoom range to the bounds of the shown genome (I.e., total length of assembly's chromosome sizes), that may be more useful than `Infinity`.

alexander-veit:
I think that omitting the second value should mean that zooming out is not restricted (i.e. `Infinity`). That's what I would expect as a user, especially since by default there are no zoom restrictions.

alexander-veit:
Only specifying one limit is now supported.

pkerpedjiev:
Really sorry for the delayed reply. A couple thoughts:

1. Zoom limits should be symmetric. So if we allow limiting the minimum we should allow limiting the maximum.

Maybe enforcing two values and having undefined indicate free zooming in that direction?

2. I like @alexpreynolds idea but we may want to make that part of another PR since it's a more complicated implementation. HiGlass doesn't really have a concept of genome sizes on the view level but each track should provide `min_pos` and `max_pos` in its tileset info. We could provide an option to enforce no zooming out beyond the highest `max_pos` of any track?

alexpreynolds:
Specifying a maximum is a great idea. Currently, one can zoom out past the true extent of the genome. To deal with this, I use a hack that checks the location on a position change. If the bounds are genome-wide, then I am likely zoomed out past the extents of the genome. So I "rubber-band" or snap the view back to the genome extents, minus 20 bases or so at the edges. Having the higlass view take care of that (if I specify genome extent as the maximum) would be a great feature to have.

alexander-veit:
Zoom limits are now symmetric. Both limits are enforced but can be `null`, which means 'no restriction'.

PR #886, opened 2020-04-02
Get local tile fetcher to work with bar tracks

alexander-veit:
## Description

The local tile fetcher was adjusted to work also with numerical data, especially the bar track.

> Why is it necessary?

The local tile fetcher did not support the bar track.

Fixes #___

## Checklist

- [x] Updated CHANGELOG.md


app/scripts/BarTrack.js
```diff
@@ -57,6 +59,7 @@ class BarTrack extends HorizontalLine1DPixiTrack {
    * Create whatever is needed to draw this tile.
    */
   initTile(tile) {
+    if (!this.initialized) return;
```
pkerpedjiev, line 62: Why not break out the stuff in the constructor into an `initialize` function and then call it instead of returning here?

alexander-veit: I tried to do it like in the `BedlikeTrack` but then I would get errors since the `HorizontalLine1DPixiTrack` did not execute completely. After fixing that it still complained since `HorizontalTiled1DPixiTrack` did not execute completely, and so on. I would have to change everything until `TiledPixiTrack`. The problem is that initializations involving the context cannot be easily moved to an `initialize` function, so in the end I couldn't get it to work.

Just returning in the `BarTrack` was the way simpler solution.

PR #890, opened 2020-04-09
Added a cheat code and bedlike improvements

pkerpedjiev:
## Description

- Added a cheat code which, if typed, toggles the header
Apr-09-2020 16-59-46

- Added `fontColor`, `fontSize` and `fillOpacity` options to the BedLikeTrack

image

- Fix track name in the config menu

image

> Why is it necessary?

Fixes #___

## Checklist

- [x] Updated CHANGELOG.md


app/scripts/HiGlassComponent.js
```diff
@@ -831,7 +835,37 @@ class HiGlassComponent extends React.Component {
...
+  typedTextHandler(event) {
+    this.typedText = this.typedText.concat(event.key);
+
+    if (this.typedText.endsWith('hgedit')) {
+      this.toggleTypedEditable();
+      this.typedText = '';
+    }
+
+    // 1.5 seconds to type the next letter
+    const TYPED_TEXT_TIMEOUT = 1500;
```
flekschas, line 854: Seems a bit long? Maybe `750` would do as well?

app/scripts/HiGlassComponent.js
```diff
@@ -831,7 +835,37 @@ class HiGlassComponent extends React.Component {
...
   }
 
+  toggleTypedEditable() {
+    this.setState({
+      typedEditable: !this.isEditable()
+    });
+  }
+
+  /** Handle typed commands (e.g. e-d-i-t) */
+  typedTextHandler(event) {
```
flekschas, line 845:
```suggestion
  typedTextHandler(event) {
    if (document.hasFocus()) return;
```

app/scripts/HiGlassComponent.js
```diff
@@ -831,7 +835,37 @@ class HiGlassComponent extends React.Component {
...
+    this.typedText = this.typedText.concat(event.key);
```
flekschas, line 846: Just out of curiosity, why not simply `this.typedText.push(event.key)`?

pkerpedjiev: No good reason.

app/scripts/HiGlassComponent.js
pkerpedjiev: I'm a slow typer ;-) But I'll change it to 750.

pkerpedjiev:
Thanks for the suggestions!

I used `concat` instead of `push` because it's two strings that are being joined. `document.hasFocus()` always seems to return true so I took it out.

I also added a change to allow users to pass in a custom PIXI import for using canvas.

PR #891, opened 2020-04-10
valueScaleMin and valueScaleMax example

nvictus:
Somewhere to point people to for doing this.

### DIFF
--- a/docs/view_config.rst
+++ b/docs/view_config.rst
@@ -54,7 +54,7 @@ The editor support a couple of keyboard shortcuts to make editing fast:
 - ``CMD (or CTRL) + Enter`` to save, apply, and finalize changes and close the modal
 - Hold ``ALT`` for 1 second to temporarily hide the modal. The modal will reappear as soon as you release ``ALT``.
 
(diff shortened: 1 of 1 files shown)


docs/view_config.rst
```diff
@@ -54,6 +54,46 @@ The editor support a couple of keyboard shortcuts to make editing fast:
 - ``CMD (or CTRL) + Enter`` to save, apply, and finalize changes and close the modal
 - Hold ``ALT`` for 1 second to temporarily hide the modal. The modal will reappear as soon as you release ``ALT``.
 
+For example, here is a useful recipe for **fixing the value scale limits** when displaying a Hi-C map using a linear color scale. First, locate the description of the heatmap track you want to edit and find the options dictionary.
```
nvictus, line 57:
```suggestion
For example, here is a recipe for **fixing heatmap value scale limits** which is useful when browsing Hi-C data with a linear color scale because of its very high dynamic range. First, locate the description of the heatmap track you want to edit and find the options dictionary.
```

PR #892, opened 2020-04-15
Retain value locks on change track type

pkerpedjiev:
## Description

- Keeps value locks when the track type is changed
- Fixes bug when value scale locks refers to non-existent tracks

> Why is it necessary?

- Value lock integrity and stability

Fixes #___

## Checklist

- [x] Updated CHANGELOG.md


pkerpedjiev:
Good point. But let's add that in another PR. I created an issue to track it here: https://github.com/higlass/higlass/issues/893

PR #898, opened 2020-04-22
Use itemRgb field for coloring bedlike tracks

pkerpedjiev:
## Description

This PR enables bedlike tracks to use the `itemRgb` field to color annotations.

See https://genome.ucsc.edu/FAQ/FAQformat.html#format1 for a description of the BED format as well as the location of the `itemRgb` field within it.

> Why is it necessary?

To follow existing standards

## Screenshot

image

Fixes #___

## Checklist

- [x] Updated CHANGELOG.md


docs/track_types.rst
```diff
@@ -46,7 +46,8 @@ column" section).
 Intervals can visually encode information using the following three ``options``:
 
 - **colorEncoding**: int or null
```
alexander-veit, line 48: int, 'itemRgb' or null

alexander-veit: Also, update the `valueColumn` description to:
```
**valueColumn**: int or null
    Use one of the bed columns (1-based) to scale the annotations along the y axis. (default `null`)
```
`none` does not work.

alexander-veit:
I did some tests with `valueColumn` and found a minor issue:
Screen Shot 2020-04-22 at 10 24 33 AM
The rectangles slightly overlap, which does not look great.

pkerpedjiev:
Thanks Alex!

> I did some tests with valueColumn and found a minor issue:

This is actually expected. The annotations have a set height. There's an option to set their height to "scaled" which will scale it according to how many are stacked. Might be good to automatically switch to that option when switching to value scaling. I'll leave that for another PR though.

docs/track_types.rst
pkerpedjiev: 👍 Good catch!

alexander-veit:
Thanks. I am not sure if we are talking about the same thing though. When I look at the last yellow rectangle, I would have expected it to be a few pixels more to the right, so that it does not overlap with the blue one (wrt `x`-direction). Is this what you meant?

pkerpedjiev:
Oh, indeed we are not talking about the same thing. I added an issue for the thing that you observed: https://github.com/higlass/higlass/issues/901

PR #904, opened 2020-04-24
Publish viewUid and trackUid with wheel events

keller-mark:
## Description

This pull request adds the properties `trackUid` and `viewUid` to the object that is published on wheel events (that have been registered via `.on('wheel', callback)` using the JS API).

> Why is it necessary?

Currently, wheel events only publish the track with which they are associated. However, this could be ambiguous if one wants to identify the track since there is no information about the parent view. To be more precise about which track a wheel event is associated with, we can also include the view UID in the event object.

Fixes #___

## Checklist

- [x] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Release notes
 
-- Fix: Add missing initializer value to `utils.reduce`
+## Next release
+
+- Add missing initializer value to `utils.reduce`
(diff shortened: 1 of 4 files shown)


manzt (approved):
Thanks for the PR! I rebased these changes on our the v2 branch and these changes will be in that release.

PR #913, opened 2020-05-10
Give warning when a malformed JSON is dropped onto HiGlass

### LINKED ISSUE #145: Give warning when a malformed JSON is dropped onto higlass
Currently the only indication that the added JSON is not valid is found in the command line.

sehilyi:
## Description

Add a panel in the viewConfig editor view that shows whether the edited json code is valid according to the definition of the higlass json schema. The failure of parsing JSON codes gives an error message and invalid configs according to the schema definition gives warning messages.

> Why is it necessary?

People can easily check whether their viewConfig is valid or not without seeing the console messages in the browser.

schema-warning

Fixes #145 

## Checklist

- [x] Updated CHANGELOG.md


app/scripts/ViewConfigEditor.js
```diff
@@ -94,7 +128,35 @@ class ViewConfigEditor extends React.Component {
...
+    if (this.state.showLog) {
+      this.hideLog();
+    } else {
+      this.showLog();
+    }
+  }
+
   render() {
+    const logMessages = this.state.logMsgs.map((d, i) => {
+      const key = `${i}-${d.type}`;
```
pkerpedjiev, line 149: These keys may not be unique.

For example, you do something and get  one  error message `1-Warning`. Then you fix that error  and get a different warning which would now also be `1-Warning` but may have a different `msg`.

pkerpedjiev:
Oh, would you mind fixing the build before merging?

sehilyi:
I was testing with a monaco-editor (https://github.com/react-monaco-editor/react-monaco-editor) which I found it to be useful in editing higlass view configs. Some advantages:
- You can see warnings and errors directly on the code with yellow and red underlines and tooltip messages based on a JSON schema.
- Suggestions on typing object keys + auto complete.
- Collapsing and expanding brackets.

monaco-editor

If you think this can be useful in addition to showing the log messages, I can try adding this editor by replacing the current 'react-simple-code-editor'.

app/scripts/ViewConfigEditor.js
sehilyi: You are right, and thanks for the explanation! I've replaced it to `d.msg`.

pkerpedjiev:
About the Monaco editor, the higlass bundle size is already very large and since the viewconf editor is meant for quick one-off edits if the Monaco editor bundle is significantly larger than the current editor's bundle, I would vote for sticking with the slimmer version.

sehilyi:
That is true, and it was my main concern with the Monaco editor as well. I will examine the size whether it is actually feasible to adopt it in higlass.

sehilyi:
I have updated a document and now tests are failing. Would this be problematic to merge?

pkerpedjiev:
No, the tests were failing on the develop branch. They should be fixed now though. I'll merge it shortly.

PR #916, opened 2020-05-17
Let plugin tracks define their own options

pkerpedjiev:
## Description

- This PR lets plugin tracks to declare their own options that are visible in the track config menu using the `optionsInfo` part of their `config`.

Example: The `higlass-arcs` track provides the `arcStyle` option: https://github.com/higlass/higlass-arcs/blob/1f68c65df24e856a993ca6d753aa78729139b8bf/src/Arcs1DTrack.js#L283

> Why is it necessary?

- So that plugin tracks can declare their own options.

Fixes #___

## Checklist

- [x] Updated CHANGELOG.md


app/scripts/SeriesListMenu.js
```diff
@@ -19,6 +19,20 @@ export default class SeriesListMenu extends ContextMenuContainer {
   getConfigureSeriesMenu(position, bbox, track) {
     const menuItems = {};
 
+    // plugin tracks can offer their own options
+    // if they clash with the default higlass options
+    // they will override them
+    const pluginOptionsInfo =
+      window.higlassTracksByType &&
+      window.higlassTracksByType[track.type] &&
+      window.higlassTracksByType[track.type].config.optionsInfo;
```
flekschas, line 28: Is `window.higlassTracksByType[track.type].config` always available? If not this could lead to issues when `window.higlassTracksByType[track.type]` is available but `window.higlassTracksByType[track.type].config` is not. Simple fix:

```
const pluginOptionsInfo =
      window.higlassTracksByType &&
      window.higlassTracksByType[track.type] &&
      window.higlassTracksByType[track.type].config &&
      window.higlassTracksByType[track.type].config.optionsInfo;
```

pkerpedjiev: It probably will always be available but it doesn't hurt to check.

PR #922, opened 2020-05-28
Horizontal multivec server-side aggregation support

keller-mark:
## Description

Adds the options `selectRowsAggregationMethod` (values: `"client"` (default) or `"server"`).

> Why is it necessary?

Allows for server-side aggregation of rows of the multivec track by POSTing the `selectRows` array and the aggregation function to use (corresponding server changes are here https://github.com/higlass/higlass-server/compare/develop...hms-dbmi:keller-mark/aggregation-post-api)

Fixes #___

## Checklist

- [x] Updated CHANGELOG.md


pkerpedjiev:
Maybe we can try to get the server changes merged in before this? Otherwise we're introducing some incompatibility?

app/scripts/TrackRenderer.js
```diff
@@ -5,6 +5,7 @@ import { zoom, zoomIdentity } from 'd3-zoom';
 import { select, event, clientPoint } from 'd3-selection';
 import { scaleLinear } from 'd3-scale';
 import slugid from 'slugid';
+import some from 'lodash/some';
```
pkerpedjiev, line 8: Do we use lodash anywhere else? I.e. do you know if it's already part of the bundle?

If not, I would suggest perhaps just implementing this function so that we're not adding another utility library.

If it is then carry on :-)

package-lock.json
```diff
@@ -13688,7 +13688,7 @@
     },
     "parse-uri": {
       "version": "1.0.0",
-      "resolved": "https://artifactory.zymergen.net:443/api/npm/zymergen-npm/parse-uri/-/parse-uri-1.0.0.tgz",
```
pkerpedjiev, line 13691: Whoops 😬

app/scripts/TrackRenderer.js
pkerpedjiev: Oh it was part of package-lock.json. Never mind.

package-lock.json
pkerpedjiev: Note to self, use `npm ci` in the tests so that we catch stuff like this.

app/scripts/TrackRenderer.js
```diff
@@ -1486,6 +1487,18 @@ class TrackRenderer extends React.Component {
       }
     }
 
+    if (
+      track.options &&
+      track.options.selectRows &&
+      track.options.selectRowsAggregationMethod === 'server' &&
```
pkerpedjiev, line 1493: We shouldn't be checking for track-specific options in `TrackRenderer`. You can set dataConfig directly in the `HorizontalMultivecTrack` instead of checking every track for these options here.

app/scripts/services/tile-proxy.js
```diff
@@ -517,7 +540,8 @@ export const tileDataToPixData = (
   zeroValueColor,
   selectedRows,
   selectedRowsAggregationMode,
-  selectedRowsAggregationWithRelativeHeight
+  selectedRowsAggregationWithRelativeHeight,
```
pkerpedjiev, line 543: The parameters for this function are getting bulky. Just planting the seed of the thought of refactoring it at some point.

app/scripts/worker.js
```diff
@@ -198,60 +267,19 @@ export function workerSetPix(
...
-              pixRowI * shape[1] + colI, // pixData index
-              d, // data point
-            );
-            pixRowI++;
-          }
-        } // end row group for
-      } // end col for
+      // Call the workerSetPixForSelectedRows helper function,
+      // which will loop over the data for us and call setPixData().
+      workerSetPixForSelectedRows(
```
pkerpedjiev, line 274: Nice, thanks for refactoring that!

package.json
```diff
@@ -66,6 +66,7 @@
     "enzyme-adapter-react-16": "^1.2.0",
     "eventemitter3": "^2.0.2",
     "genbank-parser": "^1.0.0",
+    "lodash": "^4.17.15",
```
pkerpedjiev, line 69: What's this used for?

keller-mark: Whoops, this is outdated, I will remove. I was using the `some` function originally but then removed it.

keller-mark:
No, I don't think I documented that previously. Just made this commit https://github.com/higlass/higlass/pull/922/commits/29ec8eacd92fac75a6f145e117d7fa1ca715d3a3 with some documentation.

docs/higlass_server.rst
```diff
@@ -109,6 +109,67 @@ To modify a tileset name, specify the tileset `uuid` in the URL, use the `PATCH`
 
   curl --user ${username}:${password} --request PATCH --header "Content-Type: application/json" --data '{"name":"new_name_of_tileset"}' http://localhost:8000/api/v1/tilesets/${uuid}/
 
+To retrieve a particular tile, in this case tile `1` at zoom level `6` for the tileset named by the `${uuid}` variable:
+
+.. code-block:: bash
+
+  curl localhost:8000/api/v1/tiles?d=${uuid}.6.1
+
+To retrieve many tiles at once, in this case tiles `0`, `1`, and `2` at zoom level `6` for the tileset `${uuid}`:
```
pkerpedjiev, line 118: Did  you try rendering this? I think with rst you need double backticks to get code formatting.

You can try it out like this:

```
cd docs
./serve.sh
```

keller-mark: No I did not. I just tried it and updated the formatting.

pkerpedjiev:
Oh, what about tests?

keller-mark:
Good point, just added some tests in `test/tile-proxy.js`

PR #926, opened 2020-06-06
A few fixes to the BedLikeTrack

pkerpedjiev:
## Description

- Display an error message on tiles with errors in them
- Don't try to assign a color value scale if the colorEncoding is set to `itemRgb`. `itemRgb` means that the color is present directly in the annotation
- Add an option to separate the + / - strand annotations in the bedlike track 

image

vs.

image

> Why is it necessary?

Fixes #___

## Checklist

- [x] Updated CHANGELOG.md


alexander-veit:
What is supposed to happen, when `separatePlusMinusStrands === true` and the `valueColumn` option is used? Currently this happens:
Screen Shot 2020-06-09 at 11 58 06 AM
Should the + and - strand be separate in this case?

pkerpedjiev:
Good question! `valueColumn` should override `separatePlusMinusStrands`. I'll add that to the docs.

app/scripts/TiledPixiTrack.js
```diff
@@ -740,11 +747,25 @@ class TiledPixiTrack extends PixiTrack {
         uuid: this.uuid
       });
     }
+    const errors = Object.values(this.fetchedTiles)
+      .map(
+        x =>
+          x.tileData && x.tileData.error && `${x.tileId}: ${x.tileData.error}`
```
alexander-veit, line 753: I am curious why `${x.tileId}: ${x.tileData.error}` is needed. Could you explain?

alexander-veit:
One more comment (not related to this PR): The vertical positioning of the rows seems a bit odd. They don't seem to be centered. The second row gets slightly cut off, if the track height is too small, even though there is enough space for both of them. It doesn't seem to be symmetric, since the first row is completely visible.
Screen Shot 2020-06-09 at 12 17 04 PM

app/scripts/TiledPixiTrack.js
pkerpedjiev: That just shows the error message. So if there's an error in a tile you'll something like `x.0.14: Poorly formed tile`. At the end of the call chain, we'll end up with an array of errors: `['x0.0: Tile too large', 'x.1.1: Tile malformed']` and then those will be displayed to the user.

pkerpedjiev:
> Personally, I prefer to have the test viewconfs in separate json files. In this way they also get schema checked by Travis. Currently there are several viewconfs within the tests that are invalid.

That makes sense. I'll move them out later this evening.

> The vertical positioning of the rows seems a bit odd. They don't seem to be centered.

You're right. We can fix that in a separate PR.

pkerpedjiev:
Ah, and the cursed gene annotations click test fails again. I'll try to fix that before merging. I know I promised to look into that this weekend @keller-mark. Now with this PR, it's personal ;-)

Issue #929, opened 2020-06-14
vertical divergent bar track type?


mimakaev:
It appears to be missing. 

Would be a helpful feature to have! When working with compartments a lot, it would be nice to have divergent bar charts for X and Y axis, not just on top.

pkerpedjiev:
Thanks for the heads up. This will be part of a larger PR that gets rid of the notion of "horizontal-" and "vertical-" tracks. Should be coming soon.

In the meantime, vertical divergent tracks are actually supported if you just call a vertical-bar track a "horizontal-divergent-bar". See the video below to see how to convert it:

![Jun-14-2020 13-50-03](https://user-images.githubusercontent.com/2143629/84603791-06029e00-ae46-11ea-875e-7edb483effe2.gif)

* Note that when editing the viewconf, you have to change or remove the `uid` of the track when changing its type to ensure that it rerenders.

pkerpedjiev:
Implemented in v1.10.0.

Take a look here: http://higlass.io/l/?d=Dakg-l1dTOycEp2_EgZs5A

PR #930, opened 2020-06-15
Consolidated horizontal and vertical tracks

pkerpedjiev:
## Description

- A few of the tracks we support have "horizontal-*" or "vertical-*" in their name (e.g. "horizontal-gene-annotations). This PR replaces those tracks with position-less alternatives (e.g. "gene-annotations"). The changes are reverse compatible with previous viewconfs.
- Add `trailingComma: "all"` to  `.prettierrc` because not having a trailing comma after ending parameters causes unnecessary diffs.

See https://github.com/higlass/higlass/issues/929 for a reason for this change.

Fixes #___

## Checklist

- [x] Updated CHANGELOG.md


app/scripts/HiGlassComponent.js
```diff
@@ -3890,7 +3890,6 @@ class HiGlassComponent extends React.Component {
       }
 
       if (!v.initialXDomain) {
-        console.warn('No initialXDomain provided in the view config.');
```
pkerpedjiev, line 3893: I've never once found this helpful. It's just annoying. If no initialXDomain is provided, higlass automatically zooms to the data extent.

flekschas: Nice! Blast this line of code into the void!1!

Minor comment: In the next line we're setting `initialXDomain` to `[0, 100]` which is not the data extent and can cause confusion. If the value assignment is meaningless (i.e., higlass does zoom to the data extent at some point) then I think we should fix the next line too.

app/scripts/configs/datatype-to-track-type.js
```diff
@@ -3,7 +3,26 @@ import { TRACKS_INFO } from '.';
...
+  // can a track be placed in a given orientation
+  const orientationMatches = x => {
+    // if the track's orientation matches the given orientation, then yes
+    if (x.orientation === orientation) {
+      return true;
+    }
+
+    if (
+      orientation === '1d-vertical' &&
+      x.orientation === '1d-horizontal' &&
```
flekschas, line 15: What about the other way around? `orientation` is `1d-horizontal`?

app/scripts/configs/datatype-to-track-type.js
```diff
@@ -3,7 +3,26 @@ import { TRACKS_INFO } from '.';
 export const DATATYPE_TO_TRACK_TYPE = orientation => {
   const localDatatypeToTrackType = {};
 
-  TRACKS_INFO.filter(x => x.orientation === orientation).forEach(ti => {
+  // can a track be placed in a given orientation
+  const orientationMatches = x => {
```
flekschas, line 7: What is `x` here? I assume it's not a number but the track config or something? I'd suggest using a more verbose variable name so it's easier to make sense of what the function is trying to do (testing whether a track can be in a certain orientation I assume)

app/scripts/configs/tracks-info.js
```diff
@@ -223,10 +221,12 @@ export const TRACKS_INFO = [
     }
   },
   {
-    type: 'horizontal-heatmap',
+    type: 'linear-heatmap',
+    aliases: ['horizontal-heatmap', 'vertical-heatmap'],
     datatype: ['matrix'],
     local: false,
     orientation: '1d-horizontal',
```
flekschas, line 228: this is only kept for backward compatibility right? If so, how about adding a comment so we know when we look at the file in the future.

```suggestion
    orientation: '1d-horizontal',  // only for backward compatibility
```

Even though it'd be verbose but I'd add the same comment everywhere it matches.

I just bring this up because we have so many props and I kind of lost track which one is actually still used and which one is only there for older version of glib. (e.g., I never know what `local: false` is used for; not saying we need to fix that right now)

app/scripts/TrackRenderer.js
```diff
@@ -1157,6 +1157,11 @@ class TrackRenderer extends React.Component {
           // some track types (like overlay-track don't have a track info)
           trackOrientation = TRACKS_INFO_BY_TYPE[trackDef.type].orientation;
         }
+
+        if (trackAtZoomStart.originalTrack) {
+          // this is a LeftTrackModifier track so it's vertical
```
flekschas, line 1162: I don't understand how `trackAtZoomStart.originalTrack` is related to the track orientation, specifically why we want this track to be vertical. Do you mind adding a comment?

app/scripts/configs/tracks-info.js
pkerpedjiev: Done 👍

app/scripts/TrackRenderer.js
pkerpedjiev: I just changed the line to `if (trackAtZoomStart  instanceof LeftTrackModifier)`. Hopefully that's clearer.

flekschas: Oh yeah! Now it makes a lot of sense :)

app/scripts/configs/datatype-to-track-type.js
pkerpedjiev: Renamed `x` to  `trackInfo`.

app/scripts/configs/datatype-to-track-type.js
pkerpedjiev: The default  orientation for tracks is `1d-horizontal` so if the selected orientation is `1d-horizontal` and  the  track's orientation is  `1d-horizontal` then we're ok. Otherwise we can't place the track in the selected orientation. That's a bit confusing but I think it's correct.

app/scripts/HiGlassComponent.js
pkerpedjiev: We need to assign something to `initialXDomain` otherwise all scaling and zooming will be completely broken.

pkerpedjiev: `zoomToDataExtent` is asynchronous and depends on tracks being present. If that line isn't there, then we'll end up with a useless component.

pkerpedjiev:
> I haven't tested the PR but I assume it's working. Just out of curiosity: how exactly does the track renderer now decide whether to draw a track horizontally or vertically? I saw that by default all tracks are horizontal tracks but there was something with an originalTrack property that I didn't quite understand.

Super simple. We know in where the track was placed in the viewconf (its `position`) and we know how it was defined (its `orientation`). If those don't match, then we add a `LeftTrackModifier` to convert a horizontal track to a vertical one.

Here's the relevant line: https://github.com/higlass/higlass/blob/058cec675785373302a7deb31fbe301dfc095b4c/app/scripts/TrackRenderer.js#L1454

We can still have vertical tracks, by the way. The viewport projection, for example, still has horizontal-viewport-projection and vertical-viewport-projection tracks. This is because those are complex track types that have a lot of logic that can't easily be factored out into a LeftTrackModifiable instance.

PR #936, opened 2020-06-19
Update the genbank fetcher  to read raw data

pkerpedjiev:
## Description

- Add the `text` loader in the genbank data fetcher so that the genbank file doesn't have to be specified as a url

> Why is it necessary?

- If the data is already loaded by an external application it may not be accessible at a URL

Fixes #___

## Checklist

- [x] Updated CHANGELOG.md


alexander-veit:
What kind of escaping is expected for this to work? If I open a `.gbff` file and just copy the contents into the text property, it does not work. If I use the formatted text from GenbankFetcherTests, it works.

pkerpedjiev:
Yeah, Genbank is a very picky format. You have to make sure all  of the newlines are in  place. I loaded the Genbank elsewhere and set it as part of the viewconf programmatically.

PR #949, opened 2020-07-02
Remove text encoding

pkerpedjiev:
## Description

- Removes the text/plain encoding when requesting text

> Why is it necessary?

- Because it's not necessary. It also causes problems with presigned AWS urls

Fixes #___

## Checklist

- [x] Updated CHANGELOG.md

### DIFF
--- a/app/scripts/services/tile-proxy.js
+++ b/app/scripts/services/tile-proxy.js
@@ -67,7 +67,7 @@ const throttleAndDebounce = (func, interval, finalWait) => {
 
     if (requestId && bundledRequest[requestId]) {
       bundledRequest[requestId].ids = bundledRequest[requestId].ids.concat(
-        request.ids
+        request.ids,
       );
     } else {
       requestMapper[request.id] = bundledRequest.length;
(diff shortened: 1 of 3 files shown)


app/scripts/services/tile-proxy.js
```diff
@@ -609,15 +609,19 @@ function fetchEither(url, callback, textOrJson, pubSub) {
...
+    mime = null;
   } else if (textOrJson === 'json') {
     mime = 'application/json';
   } else {
     throw new Error(`fetch either "text" or "json", not "${textOrJson}"`);
   }
-  const headers = { 'Content-Type': mime };
+  const headers = {};
+
+  if (mime) {
```
pkerpedjiev, line 622: This is the only chane

app/scripts/services/tile-proxy.js
```diff
@@ -609,15 +609,19 @@ function fetchEither(url, callback, textOrJson, pubSub) {
   requestsInFlight += 1;
   pubSub.publish('requestSent', url);
 
-  let mime;
+  let mime = null;
   if (textOrJson === 'text') {
-    mime = 'text/plain';
+    mime = null;
```
flekschas, line 614: Seems unnecessary because you initialize `mime` with `null`.

Just curious, why is the `'text/plain'` mime type harmful?

flekschas: Nevermind, I just read your PR description, which answers my question :)

I think the entire if statement can be removed

PR #798, opened 2019-11-05
Clarify fileUrl data source in viewconfs

pkerpedjiev:
## Description

This PR updates the docs to clarify that we don't provide a `fileUrl`-compatible server out of the box.

Fixes #___

## Checklist

- [x] Set proper GitHub labels (e.g. v1.6+, ignore if you don't know)
- [ ] Unit tests added or updated
- [ ] Documentation added or updated
- [ ] Example(s) added or updated
- [ ] Update schema.json if there are changes to the viewconf JSON structure format
- [ ] Screenshot for visual changes (e.g. new tracks or UI changes)
- [x] Updated CHANGELOG.md

PR #950, opened 2020-07-06
Update fileUrl and filetype track properties

keller-mark:
## Description

This allows the `fileType`, `coordSystem` and `indexUrl` properties to be set along with `fileUrl`.

Also, I updated the documentation by copying the text from https://github.com/higlass/higlass/pull/798

> Why is it necessary?

Previously, the `filetype` property was deleted from the track config in `app/scripts/HiGlassComponent.js` line 3306.

In some places it looks like `filetype` was changed to `fileType`. Here I updated this in additional places where it appeared.

When using `fileUrl` with bigwigs stored on cistrome DB, we may want to provide an `indexUrl` to a corresponding JSON metadata location for the bigwig file, which can be returned as tileset_info by the server.

Fixes #___

## Checklist

- [x] Updated CHANGELOG.md


pkerpedjiev:
Hey Mark, I think this is fine. Before putting down the green check mark I wanted to quickly ask how you feel about moving `fileUrl` and `fileType` to the `data` section like we do with other data loaders?

See http://docs.higlass.io/view_config.html#genbank-files for one example or https://github.com/higlass/higlass-pileup/ for another.

I know it was the way it is before your modifications but I think you're the first actual user of this interface so now would be a good time to change it.

keller-mark:
That sounds good to me, I will try it. Should those properties be `data.url` and `data.type` then  / should they lose the "file" prefix when they are moved to the data section?

pkerpedjiev:
I would make it something like this:

```
{
  type: 'file-proxy',
  proxy: 'https://higlass.io/api/v1',
  url: 'https://my.bigwig.url',
  filetype: 'bigwig'
}
```

keller-mark:
I made that change but kept server as `server` rather than `proxy`, since after registering the fileUrl tileset, server and tilesetUid are used normally.

PR #959, opened 2020-07-22
Allow tracks to modify their width and height

alexander-veit:
## Description

This PR enables tracks to modify their height and width by publishing 'trackDimensionsModified'.

> Why is it necessary?

This is necessary for the Transcripts track which needs to adjust its own height when the number of shown transcripts changes.

Fixes #___

## Checklist

- [ ] Updated CHANGELOG.md


app/scripts/HiGlassComponent.js
```diff
@@ -4415,6 +4422,33 @@ class HiGlassComponent extends React.Component {
     );
   }
 
+  /**
+   * Handle trackDimensionsModified events
+   * settings.trackId = id of the track
+   * settings.height = new height of the track or undefined if current height should remain
+   * settings.width = new width of the track or undefined if current width should remain
+   */
+  trackDimensionsModifiedHandler(settings) {
```
pkerpedjiev, line 4431: Track uids are not guaranteed to be globally unique. You need a view / track uid pair to uniquely identify a track.

alexander-veit: Oh yes, of course. Thanks. Is there a simple way to get the current viewId from within a (plugin) track?

pkerpedjiev: Unfortunately, I don't think there is. You could perhaps add it to the context here: https://github.com/higlass/higlass/blob/9ceffd59cd75dff315cf67e33185da185b8c7419/app/scripts/TrackRenderer.js#L1491 but I'm not sure that it's even present in TrackRenderer.

Another possibility is to pass a callback into the track object that lets it change its size. See the viewport projection tracks as an example of this:

https://github.com/higlass/higlass/blob/9ceffd59cd75dff315cf67e33185da185b8c7419/app/scripts/TrackRenderer.js#L1621

alexander-veit: Thanks. It turns out the TrackRenderer knows about it. It just wasn't used so far.

PR #960, opened 2020-07-29
Fix server-side aggregation with dynamic selectRows option bug

keller-mark:
## Description

This pull request adds logic to update the HorizontalMultivecTrack `this.dataFetcher` when `dataConfig.options.aggGroups` has changed due to an updated view config with updated `track.options.selectRows` property, when using the server-side aggregation option. This also adds a call to `this.fetchNewTiles` after creating the new dataFetcher.

This pull request also fixes a bug encountered when using the `data.url` track option (previously known as `fileUrl`) with server-side aggregation: the HorizontalMultivecTrack's dataFetcher was not updating after registering the URL and receiving a `tilesetUid`.

Also - the pre-commit linting / formatting hook changed many lines of the `test/HorizontalMultivecTests.js` that were unrelated to the test that I wrote. I cannot tell if these are regression from good formatting or progress toward good formatting (if they are bad changes I will manually revert).

> Why is it necessary?

In testing out higlass v1.10.2 in the Cistrome Explorer, I discovered these two bugs that I introduced that prevent the HorizontalMultivecTrack from updating in response to an updated view config (with updated track options for aggregation).

I previously added code in #922 to set `this.dataFetcher` and `this.selectedRowsOptions` in the HorizontalMultivecTrack constructor, which prevents these variables from updating in response to changes to updated track options ([here](https://github.com/higlass/higlass/blob/29a30384f19f9dd10ae2e7355e96e963e3a61cca/app/scripts/HorizontalMultivecTrack.js#L35) and [here](https://github.com/higlass/higlass/blob/29a30384f19f9dd10ae2e7355e96e963e3a61cca/app/scripts/HorizontalMultivecTrack.js#L41)).

Fixes #___

## Checklist

- [x] Updated CHANGELOG.md

### DIFF
--- a/app/scripts/HorizontalMultivecTrack.js
+++ b/app/scripts/HorizontalMultivecTrack.js
@@ -1,4 +1,5 @@
 import { format } from 'd3-format';
+import equal from 'fast-deep-equal';
 
 import HeatmapTiledPixiTrack from './HeatmapTiledPixiTrack';
 import DataFetcher from './DataFetcher';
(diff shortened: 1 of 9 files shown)


app/scripts/HorizontalMultivecTrack.js
```diff
@@ -1,4 +1,5 @@
 import { format } from 'd3-format';
+import equal from 'fast-deep-equal';
```
pkerpedjiev, line 2: Grumble grumble, another import.

FWIW, unless speed is of the utmost importance (i.e. the operation is being performed multiple times on every redraw) `JSON.stringify(obj1) === JSON.stringify(obj2)`  is usually fast enough. God help you if you have a circular structure though.

keller-mark: I have updated this to remove the import

keller-mark:
It looks like some tests are failing due to the resgen.io server being down

```
29 07 2020 15:42:11.430:ERROR [Chrome 84.0.4147 (Linux 0.0.0) | BedLikeTrack |, inline annotations | checks that the rectangles are strand separated or not]: 'Could not fetch https://resgen.io/api/v1/gt/paper-data/tilesets/?limit=100&dt=gene-annotation', TypeError{}
```

but I will merge since the tests passed previously https://travis-ci.org/github/higlass/higlass/builds/712768947

PR #985, opened 2020-09-25
Categorical mouseover appearance and first-/last-row bug patch

alexpreynolds:
## Description

1) I customize `getMouseOverHtml` in `HorizontalMultivecTrack` so that the mouseover tooltip has a similar appearance to the `higlass-multivec` track used for epilogos tracks. This only affects viewconfs which have `heatmapValueScaling` defined in the track's `options`, where it is set to `categorical`. The tooltip is otherwise presented in the same way as before.

2) A mouseover event on the first row of a multivec track would return no value for `row_infos`. This is because the index of the first row (0) would return `false` in the test `else if (selectedRowIndex)`, which skips adding the row information. I instead test if `selectedRowIndex` is within the bounds of the `row_infos` array, to add that value.

3) Mouseover on the multivec track below the last row would return `undefined` color and label values. I test if this case occurs and return early, if so.

> Why is it necessary?

To bring the tooltip appearance more in line with what the `higlass-multivec` track displays for epilogos/categorical tracks, and to fix two small mouseover bugs.

Fixes #___

## Checklist

- [ ] Updated CHANGELOG.md

### DIFF
--- a/app/scripts/HorizontalMultivecTrack.js
+++ b/app/scripts/HorizontalMultivecTrack.js
@@ -372,7 +372,10 @@ export default class HorizontalMultivecTrack extends HeatmapTiledPixiTrack {
       let rowInfo = '';
       if (this.options.selectRows && !Array.isArray(selectedRowItem)) {
         rowInfo = this.tilesetInfo.row_infos[selectedRowItem];
-      } else if (selectedRowIndex) {
+      } else if (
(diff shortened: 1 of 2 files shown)


pkerpedjiev (approved):

pkerpedjiev:
Thanks Alex! The tests are a mess right now so I'll just merge it sans.

PR #991, opened 2020-10-23
Flekschas/export overlays as svg

flekschas:
## Description

Implement SVG export of overlays.

Example screenshot from http://localhost:8080/apis/svg.html?/viewconfs/overlay-tracks.json:

Screen Shot 2020-10-23 at 10 03 23 AM

> Why is it necessary?

To see overlays in the exported SVG

## Checklist

- [x] Updated CHANGELOG.md

### DIFF
--- a/app/scripts/OverlayTrack.js
+++ b/app/scripts/OverlayTrack.js
@@ -86,7 +86,7 @@ const drawRectWithPositionedBorder = (
     colors.strokeOpacity = 0;
     colors.strokeWidth = 1;
 
-    outline.positions.forEach(pos => {
+    outline.positions.forEach((pos) => {
       if ((pos === 'top' && !isVertical) || (pos === 'left' && isVertical)) {
         drawRect(
           xPos - outlineWidth,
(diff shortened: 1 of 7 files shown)


app/scripts/OverlayTrack.js
```diff
@@ -299,12 +389,13 @@ class OverlayTrack extends PixiTrack {
...
               stroke,
               outline,
+              options,
             ),
           );
         }
 
         if (op.orientation === '1d-vertical' || op.orientation === '2d') {
-          this.options.extent.forEach((extent) =>
+          this.options.extent.forEach(extent =>
```
pkerpedjiev, line 398: Which version of prettier are you using? The default (in v2) should be to add parens around anonymous function arguments.

app/scripts/OverlayTrack.js
```diff
@@ -234,6 +322,8 @@ class OverlayTrack extends PixiTrack {
       height = minHeight;
     }
 
+    options.isVertical = true;
```
pkerpedjiev, line 325: Not sure why this is here but I presume it's intentional?

pkerpedjiev (commented):
The only reason I'm not approving is the prettier thing. Should be easy to fix.

app/scripts/OverlayTrack.js
flekschas: Yes it is. it's part of the function to draw a vertical annotation

app/scripts/OverlayTrack.js
flekschas: Forgot to do `npm i`. Thanks for the catch

pkerpedjiev (approved):

PR #997, opened 2020-12-01
Add a zoomToGene API

sehilyi:
## Description

> What was changed in this pull request?

This PR adds a `zoomToGene` API that allows zooming a HiGlass view to a location near a certain gene. The location of gene is searched based on view specs: `autocompleteServer` and `autocompleteId`.

```js
hgApi.zoomToGene(
  'view-1', // view id
  'MYC',    // gene name
  2000      // duration of animated transition
);
```

> Why is it necessary?

This allows to reposition the center of a HiGlass view using a gene name, similar to a `zoomTo` API that instead uses absolute positions.

Fixes #___

## Checklist

- [ ] Set proper GitHub labels (e.g. v1.6+, ignore if you don't know)
- [x] Unit tests added or updated
- [x] Documentation added or updated
- [ ] Example(s) added or updated
- [ ] Update schema.json if there are changes to the viewconf JSON structure format
- [ ] Screenshot for visual changes (e.g. new tracks or UI changes)
- [x] Updated CHANGELOG.md

PR #999, opened 2020-12-02
Add a suggestGene API

sehilyi:
## Description

This PR adds an API, `suggestGene`, that returns the list of top matching genes for a given keyword (e.g., 'MY'). The suggestions are based on view specs: `autocompleteServer` and `autocompleteId`.

```js
hgApi.suggestGene('aa', 'MY', (suggestions) => {
  if(suggestions && suggestions.length > 0) {
    console.log('Gene suggested', suggestions[0].geneName);
    console.log('Chromosome', suggestions[0].chr);
    console.log('Start position', suggestions[0].txStart);
    console.log('End position', suggestions[0].txEnd);
  }
});
```

> Why is it necessary?

Together with the `zoomToGene` API (#997), this can be useful when we want to support searching for genes on the interface that are placed outside of HiGlass views.

Fixes #___

## Checklist

- [x] Updated CHANGELOG.md

### DIFF
--- a/app/scripts/HiGlassComponent.js
+++ b/app/scripts/HiGlassComponent.js
@@ -4054,7 +4054,7 @@ class HiGlassComponent extends React.Component {
     this.setCenters[viewUid](centerX, centerY, k, false, animateTime);
   }
 
-  zoomToGene(viewUid, gene, animateTime) {
+  zoomToGene(viewUid, geneName, animateTime) {
     if (!(viewUid in this.setCenters)) {
       throw Error(
         `Invalid viewUid. Current uuids: ${Object.keys(this.setCenters).join(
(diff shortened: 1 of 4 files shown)


app/scripts/HiGlassComponent.js
```diff
@@ -4075,53 +4075,82 @@ class HiGlassComponent extends React.Component {
...
-              this.pubSub,
-            );
-          } else {
-            console.warn(`Couldn't find the gene symbol: ${geneSymbol}`);
-          }
-        }
+      .then((suggestions) => {
+        callback(
+          suggestions.map((d) => {
+            // change to use simpler key names just to make it easier to use
```
pkerpedjiev, line 4144: I don't think this is a good idea. It's introducing an inconsistency between the data returned from the server API and client API.

Not worth making it "easier to use", IMO. Also, I don't necessarily agree that it's easier to use. `txStart` and `geneName` convey more information than `start` and `gene`.

pkerpedjiev (commented):
Just one comment on renaming the return values.

app/scripts/HiGlassComponent.js
sehilyi: Thanks for pointing this out. I guess I was trying to over-simplify the variable names, and I agree that it is not a good idea to introduce the inconsistency. I've renamed the variables to use the names as they were.

PR #1004, opened 2021-01-20
Allow implementing plugin viewport tracks

sehilyi:
## Description

This PR is for allowing to implement plugin viewport projection tracks (similar to `viewport-projection-center`). Plugin tracks' `config.projection` option is checked to determine if callback functions should be added (e.g., `setDomainsCallback`). 

Extending `SVGTrack` in a plugin track and setting `config.projection` to `true` allows to sync its genomic coordinates with other HiGlass views.

> Why is it necessary?

We want to implement a custom viewport track for circular layouts.

Fixes #___

## Checklist

- [x] Updated CHANGELOG.md


app/scripts/HiGlassComponent.js
```diff
@@ -3074,7 +3074,8 @@ class HiGlassComponent extends React.Component {
     if (
       track.type === 'viewport-projection-center' ||
       track.type === 'viewport-projection-horizontal' ||
-      track.type === 'viewport-projection-vertical'
+      track.type === 'viewport-projection-vertical' ||
+      (track.options && track.options.addCallbacks)
```
pkerpedjiev, line 3078: This might be a problem because a track's options are set by the user and if they remove the `addCallbacks` section then it'll stop working. You could maybe use the track's datatype setting (see `tracks-info.js`)?

pkerpedjiev (commented):
Just the comment about using `addCallbacks` in `track.options`.

app/scripts/HiGlassComponent.js
sehilyi: Thank you for this suggestion! I've changed to use `config.projection` instead, which is not visible to users.

PR #1018, opened 2021-03-12
Added a docstring for  API function

pkerpedjiev:
## Description

- Updated the docstring for `getTrackObject`

> Why is it necessary?

Fixes #___

## Checklist

- [ ] Updated CHANGELOG.md

### DIFF
--- a/app/scripts/api.js
+++ b/app/scripts/api.js
@@ -668,7 +668,7 @@ const createApi = function api(context, pubSub) {
        * Set or get an option.
        * @param   {string}  key  The name of the option you want get or set
        * @param   {*}  value  If not `undefined`, `key` will be set to `value`
-       * @return  {object}  When `value` is `undefined` the current value of
+       * @return  {obj}  When `value` is `undefined` the current value of
        *   `key` will be returned.
        */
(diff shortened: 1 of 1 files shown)


app/scripts/api.js
```diff
@@ -664,7 +668,7 @@ const createApi = function api(context, pubSub) {
        * Set or get an option.
        * @param   {string}  key  The name of the option you want get or set
        * @param   {*}  value  If not `undefined`, `key` will be set to `value`
-       * @return  {[type]}  When `value` is `undefined` the current value of
+       * @return  {object}  When `value` is `undefined` the current value of
```
sehilyi, line 671: Just curious: Is there any difference between `object` and  `obj`?

sehilyi (approved):
Looks good!

app/scripts/api.js
pkerpedjiev: No, and that notation has no functional significance. It's just to inform the user. With that said, consistency is nice so I'll update it to match the other return value.

```suggestion
       * @return  {obj}  When `value` is `undefined` the current value of
```

PR #1038, opened 2021-07-26
Remove cwise

### LINKED ISSUE #988: Get rid of cwise
`cwise` depends on an old version of `static-eval` which has a pretty nasty security vulnerability. We also don't use `cwise` very extensively so this should be a small change to make.

alexander-veit:
## Description

This PR removes the dependency on the `cwise` package.

> Why is it necessary?

Fixes #988

## Checklist

- [x] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Next
 
+- Remove dependency on `cwise`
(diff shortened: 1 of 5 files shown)


pkerpedjiev (approved):
I don't know exactly what cwise does but I"ll trust that this code is correct.

PR #1047, opened 2021-08-27
Remove bootstrap from the genome position search box

pkerpedjiev:
## Description

- Removed React bootstrap from the genome position search box

> Why is it necessary?

- React bootstrap is a pain to include and clobbers all styles. It would be better if we got rid of it.

Before:

image

After:

image

Fixes #___

## Checklist

- [ ] Updated CHANGELOG.md


app/scripts/icons.js
```diff
@@ -369,3 +371,18 @@ export const svgArrowheadDomainsIcon = parser.parseFromString(
   arrowHeadDomainsStr,
   'text/xml',
 ).documentElement;
+
+export function SearchIcon({ theStyle }) {
+  return (
+    <svg
+      styleName={theStyle}
```
flekschas, line 378: What is the `styleName` property used for? Should this be `className` or `style` or is this a special property?

app/scripts/icons.js
```diff
@@ -369,3 +371,18 @@ export const svgArrowheadDomainsIcon = parser.parseFromString(
   arrowHeadDomainsStr,
   'text/xml',
 ).documentElement;
+
+export function SearchIcon({ theStyle }) {
```
flekschas, line 375: How about calling the property `className` so it's clearer what this is support to be?

app/scripts/GenomePositionSearchBox.js
```diff
@@ -853,14 +841,11 @@ class GenomePositionSearchBox extends React.Component {
...
-        <button
+        <SearchIcon
           onClick={this.buttonClick.bind(this)}
-          styleName={classNameButton}
-          type="button"
-        >
-          <Glyphicon glyph="search" />
-        </button>
-      </FormGroup>
+          theStyle="multitrack-header-icon"
```
flekschas, line 846: I'd change this to `className`

flekschas (commented):
Looks good! 🎉 I'll give it a practical test in an hour but left some minor comments.

flekschas:
One small thing I found during testing: the search icon doesn't look consistent with the other icons:

1. It's darker than the other icons
2. On hover it's not changing to white

Aug-29-2021 15-50-46

My guess is that it's `fill` property isn't set to `currentColor`.

Apart from that I would like to test the tileset adding modal but higlass.io is currently down so I can't load any tilesets. Before merging I'd recommend testing that modal once higlass.io is back online.

app/scripts/icons.js
pkerpedjiev: It's how we pass style names using `react-css-modules`. Tbh, it feels a bit dated but that's what we use in other places as well so I'm not too keen on changing it just yet.

app/scripts/icons.js
pkerpedjiev: I'm afraid that that'll interfere with other functionality. `styleName` is how we pass in styles in other components.

pkerpedjiev:
Updated the normal color to the standard grey and the mouseover color to white 👍 Thanks for the suggestion!

PR #1057, opened 2021-10-03
Nan as zero

pkerpedjiev:
## Description

- Added an option that lets line tracks treat NaN values as zeros. This lets it render an unbroken line.

With the option on:

image

With the option off:

image

> Why is it necessary?

Fixes #___

## Checklist

- [x] Updated CHANGELOG.md


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Release notes
 
 - Remove react-bootstrap from the GenomePositionSearchBox
+- Added the `nanAsZero` option for `horizontal-line` tracks
```
flekschas, line 4: Does this work for vertical-line track as well? And only line track or also bar tracks?

app/scripts/HorizontalLine1DPixiTrack.js
```diff
@@ -167,7 +167,11 @@ class HorizontalLine1DPixiTrack extends HorizontalTiled1DPixiTrack {
 
     for (let i = 0; i < tileValues.length; i++) {
       const xPos = this._xScale(tileXScale(i));
-      const yPos = this.valueScale(tileValues[i] + offsetValue);
+      let yPos = this.valueScale(tileValues[i] + offsetValue);
+
+      if (Number.isNaN(tileValues[i]) && this.options.nanAsZero) {
```
flekschas, line 172: How about combining this to

```js
const yValue = this.options.nanAsZero && Number.isNaN(tileValues[i]) ? 0 : tileValues[i];
const yPos = this.valueScale(yValue + offsetValue);
```

While functionally equivalent, I personally like more as it signals to the developer: yValue and yPos are two fixed variables that cannot be changed later on. While with `let` the value could be changed elsewhere and so one would have to example the entire file.

app/scripts/Tiled1DPixiTrack.js
```diff
@@ -243,6 +243,7 @@ class Tiled1DPixiTrack extends TiledPixiTrack {
 
     const minimumsPerTile = visibleAndFetchedIds
       .map((x) => this.fetchedTiles[x])
+      .filter((tile) => tile.tileData.dense)
```
flekschas, line 246: Huh, this seems to be unrelated to the `NaN` support. What's this doing?

app/scripts/TiledPixiTrack.js
```diff
@@ -518,7 +518,7 @@ class TiledPixiTrack extends PixiTrack {
     return this.scale.maxRawValue;
   }
 
-  initTile(/* tile */) {
+  initTile(tile) {
```
flekschas, line 521: What's going on here?

flekschas (commented):
Thanks for the PR.

There were 3 changes I couldn't interpret. Other the PR looks good 👍

CHANGELOG.md
pkerpedjiev: It also works for vertical line tracks. The `horizontal` is a relic of the past that I used out of habit. `line` works as well. The vertical version uses the same code.

See here: https://github.com/higlass/higlass/blob/develop/app/scripts/TrackRenderer.js#L1581

app/scripts/TiledPixiTrack.js
pkerpedjiev: It was commented out to shut up the linter because `tile` wasn't being used before. Now it's being used.

app/scripts/Tiled1DPixiTrack.js
pkerpedjiev: Oh, good question! The test code uses the local tile fetcher which returns tile data synchronously. Somewhat paradoxically, this causes the track to try to render itself before the tile is completely filled in and throws an error if this check isn't there.

app/scripts/HorizontalLine1DPixiTrack.js
pkerpedjiev: I like it! Thanks for the suggestion. Just included it and pushed a change.

PR #1072, opened 2021-12-01
Prevent erroneous lines on 2D tracks by clearing stale mouse position data

### LINKED ISSUE #814: Multivec track shows horizontal crosshair when mouse pointer is not over multivec track
Using v1.7.2 of higlass, I have a `horizontal-multivec` track along with a few other non-multivec tracks (gene annotations, `horizontal-point`, and chromosome tracks). 

When the mouse pointer moves over a non-multivec track, the multivec track behaves as though the vertical coordinate of the mouse pointer is also on the multivec track, which enables the horizontal line of the crosshair:

Crosshair

What I would expect is that, when the multivec is part of the top or bottom of a view, it would only show the horizontal crosshair if the pointer is within the bounds of the multivec track. 

(Likewise, if the bug is consistent between orientations, a multivec that is in the left or right portion of a view would not show the vertical crosshair, unless the pointer is within the bounds of the multivec.)

alexpreynolds:
## Description

When the `showMousePosition` option is set to `true` and the mouse pointer moves over a non-2d track, x and y values are published internally that can cause another 2d track to draw nonsensical vertical or horizontal lines. This test clears the relevant value so that it cannot be used to render such lines.

> Why is it necessary?

Fixes #814

## Checklist

- [ ] Updated CHANGELOG.md

### DIFF
--- a/app/scripts/utils/show-mouse-position.js
+++ b/app/scripts/utils/show-mouse-position.js
@@ -119,6 +119,22 @@ const showMousePosition = (
       y = event.isFrom2dTrack ? event.dataY : event.dataX;
     }
 
+    // If a track is not 2d and is horizontal, the event.dataY
+    // value of the mouse pointer location should not be allowed
+    // to impact how other tracks globally render their mouse
+    // position (see https://github.com/higlass/higlass/issues/814).
+    // Likewise, if a track is not 2d and is vertical, the
+    // event.dataX value should not be used by other tracks for
+    // position reporting.
+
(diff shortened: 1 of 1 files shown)


pkerpedjiev:
Looks good! Gotta fix the dang tests (not your fault). Let me try to do that over the weekend and then we can merge it. Unless it's urgent.

alexpreynolds:
Not urgent at all, just wanted to scratch a two year old itch!

PR #1078, opened 2022-02-03
feat: check-in schema generated from higlass_schema

manzt:
## Description

The file checked in was generated via:

```bash
pip install git+https://github.com/higlass/higlass-schema
python -m higlass_schema.main > app/schema.json
```

If the changes are desirable, we can discuss how to add `higlass_schema` as a dependency to this repo.

This PR introduces `schema.json` generated from [`higlass/higlass-schema`](https://github.com/higlass/higlass-schema). It replaces the old schema with one that is generated automatically from python class definitions. 

 The initial python code was bootstrapped via a [code-generation tool](https://pydantic-docs.helpmanual.io/datamodel_code_generator/) which used the `schema.json` from this repo as input. Classes were modified manually to better reflect the types described here. 

> Why is it necessary?

This is a step towards keeping the higlass client and any python API in sync with minimal manual refinement. The idea is that `higlass/higlass-schema` serves as the "source of truth" and schema changes should occur there.

## Checklist

- [ ] Updated CHANGELOG.md

### DIFF
--- a/app/schema.json
+++ b/app/schema.json
@@ -2,524 +2,913 @@
   "$schema": "http://json-schema.org/draft-07/schema#",
   "$id": "https://higlass.io/#viewconf",
   "title": "HiGlass viewconf",
+  "description": "Root object describing a HiGlass visualization.",
   "type": "object",
-  "additionalProperties": false,
-  "required": [],
   "properties": {
     "editable": {
-      "type": "boolean",
-      "default": true
+      "default": true,
+      "type": "boolean"
     },
     "viewEditable": {
-      "type": "boolean",
-      "default": true
+      "default": true,
+      "type": "boolean"
     },
     "tracksEditable": {
-      "type": "boolean",
-      "default": true
+      "default": true,
+      "type": "boolean"
     },
-    "zoomFixed": { "type": "boolean" },
-    "compactLayout": { "type": "boolean" },
-    "exportViewUrl": { "type": "string" },
+    "zoomFixed": {
+      "type": "boolean"
+    },
+    "compactLayout": {
+      "type": "boolean"
+    },
+    "exportViewUrl": {
+      "type": "string"
+    },
     "trackSourceServers": {
+      "minItems": 1,
       "type": "array",
-      "items": { "type": "string" },
-      "minLength": 1
-    },
-    "zoomLocks": { "$ref": "#/definitions/locks/zoomLocks" },
-    "locationLocks": { "$ref": "#/definitions/locks/locationLocks" },
-    "valueScaleLocks": { "$ref": "#/definitions/locks/valueScaleLocks" },
-    "views": {
-      "type": "array",
-      "items": { "$ref": "#/definitions/view" },
-      "minLength": 1
-    },
-    "chromInfoPath": { "type": "string" }
-  },
-  "definitions": {
-    "data": {
-      "type": "object",
-      "properties": {
-        "url": { "type": "string" },
-        "server": { "type": "string" },
-        "filetype": { "type": "string" },
-        "type": { "type": "string" },
-        "tilesetInfo": { "type": "object" },
-        "children": { "type": "array" },
-        "tiles": { "type": "object" }
+      "items": {
+        "type": "string"
       }
     },
-
-    "locks": {
-      "slug": { "type": "string" },
-
-      "locksByViewUid": {
-        "type": "object",
-        "additionalProperties": false,
-        "patternProperties": {
-          ".": { "$ref": "#/definitions/locks/slug" }
+    "views": {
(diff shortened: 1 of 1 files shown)


pkerpedjiev:
Looks good. I didn't go through the schema line by line but it looks like a reasonable addition. A few questions:

1. Did you happen to try running the schema on any of the viewconfs in test/view-configs?
2. It might be worth trying to run in on a viewconf in a plugin track to see how it behaves: 

https://github.com/higlass/higlass-pileup/blob/master/src/index.html#L61

Do you know if it's possible to do something like subclass a schema? I'm thinking in terms of plugin tracks it would be nice if they could take the existing higlass schema, add whatever they need and then use that? I presume it would just be a matter of subclassing the pydantic class and regenerating the schema?

manzt:
> Did you happen to try running the schema on any of the viewconfs in test/view-configs?

I just ran the test suite (sucessfully) locally. I can go through some of the view configs.

2. It might be worth trying to run in on a viewconf in a plugin track to see how it behaves:

I'll try that out.

> I presume it would just be a matter of subclassing the pydantic class and regenerating the schema?

So I ran into something similar (for different reasons) that I need for `hg`. I wanted my derived classes to have additional python methods not implemented on the core classes. I think I found a solution where we can expose this behavior via generics and `pydantic.generics.GenericModel`. 

https://github.com/higlass/higlass-schema/pull/1/files

We can make the schema generic over different properties, and any concrete subclassing of the generic config will extend the schema.

```python
from typing import Literal
from higlass_schema as hgs

class MyCustomTrack(hgs.BaseTrack[Literal["a-custom-track"]):
  customField: str = "I'm a custom field"

hgs.Viewconf.schema_json() # base schema, "bounds" on generic types are used to generate schema
hgs.Viewconf[hgs.View[Union[hgs.Track, MyCustomTrack]].schema_json() # explicit extension
```

manzt:
Let me know if that makes any sense or if you'd want to hop on a call to discuss anything.

pkerpedjiev:
Yeah, this makes sense. Thank you! I'm thinking that plugin repos could have a small chunk of Python code that extends the schema and higlass-schema and generates their own.

Mixing Python and JS is not ideal but as far as I know there's no JS equivalent to pydantic?

manzt:
Just opened a PR to add a simple CLI for checking schemas:

- https://github.com/higlass/higlass-schema/pull/3

> Yeah, this makes sense. Thank you! I'm thinking that plugin repos could have a small chunk of Python code that extends the schema and higlass-schema and generates their own.

In that PR I expanded the Track definition to include `BaseTrack` which is a catch-all type for all non-`CombinedTrack`/`EnumTrack`/`HeatmapTrack/`. It's a more flexible type since I allowed extra fields on the model. 

TL;DR - with this change, the pile-up schema you shared is valid but all the unrecognized tracks end up being `BaseTrack` in python.

```python
from typing import Union
import higlass_schema as hgs

hgs.Viewconf[
	hgs.View[hgs.EnumTrack, hgs.CombinedTrack, hgs.HeatmapTrack]
].parse_file("./pileup.json") # fails because tracks aren't one of enum/combined/heatmaps

c = hgs.Viewconf[
	hgs.View[hgs.EnumTrack, hgs.CombinedTrack, hgs.HeatmapTrack, hgs.BaseTrack]
]parse_file("./pileup.json") # success!

track = c.views[0].tracks.top[0] # hgs.BaseTrack
print(track.type) # type == str, "chromosome-labels"
print(track.filetype) # type == Any, "chromsizes-tsv"  
```

By adding the `BaseTrack` to the root generic bound, the latter behavior is the default:

```python
hgs.Viewconf.parse_file("./pileup.json") # ok!
```

pkerpedjiev:
That seems reasonable.

By the way, why do you use the name `EnumTrack`? From what I can tell, that's meant to reference tracks that are in the base higlass distribution?

manzt:
> By the way, why do you use the name EnumTrack? From what I can tell, that's meant to reference tracks that are in the base higlass distribution?

I bootstrapped `schema.py` with a code generation tool that took the names from `schema.json` directly...

https://github.com/higlass/higlass/blob/274ebb24a95c125cfa2c6a9ed90914f81c3f0728/app/schema.json#L407

A track for base higlass is defined as `anyOf` `enum_track`, `heatmap_track`, `combined_track`, `independent_viewport_projection_track`

manzt:
Schema tests had been turned off for a while. Good thing we didn't merge. Need to have a closer look.

pkerpedjiev:
Anything to be done here on my end?

manzt:
Nope! I just need to spend some time and look at it.

manzt:
Closing as stale.

PR #1085, opened 2022-02-16
Enable click events on tracks and custom modals

alexander-veit:
## Description

Tracks and plugin tracks can now implement `onMouseClick()`, which allows them to react to click events on the corresponding track. Furthermore they can use the method `showCustomTrackDialog` to open modals and display custom data in them (e.g. in response to a click on the track).

> Why is it necessary?

I want to be able to display additional data in a modal when a certain part of a track is clicked. `showCustomTrackDialog` expects a react component with props as input. That react component can live in the plugin track repository. This makes it very flexible which the data we can show in that modal and how we present it.

Here is an example (implemented in a plugin track):
```
onMouseClick(){
  if(!this.mouseClickData) return;

  const chr = this.mouseClickData.chrName;
  const pos = this.mouseClickData.from - this.mouseClickData.chrOffset;
  
  const props = {
    chr: chr,
    pos: pos,
    dataFetcher: this.variantDetailFetcher // the data shown will be loaded from an external source
  }

  const data = {
    title: `Variant ${chr}:${format(',')(pos)}`,
    bodyComponent: VariantDetails, // this is a react component
    bodyProps: props
  };
  this.showCustomTrackDialog(data);
}
```
Here is a demo:
https://gyazo.com/9bfe2138895c3f414ed631c2ded706e0

Fixes #___

## Checklist

- [x] Updated CHANGELOG.md


alexander-veit:
I know that there is already the `app.click` event that you use for the annotation tracks, but as far as I can tell this only works when clicking on a pixi graphics object (using its `mouseup` method). Since I need to display a lot of elements that are clickable, I can't create separate graphics object for each of them. Therefore, I capture track-wide clicks and figure out afterwards, if something of interest has been clicked (like we do it for `getMouseOverHtml`).

pkerpedjiev:
What happens if two tracks which are overlayed and try to display a custom dialog.

Also would be good to update the docs with this functionality.

alexander-veit:
Sorry, just coming back to this now. I added a paragraph to the docs.

If two tracks are overlayed and you click on a region where both tracks open a dialog, only one would open. I don't know if there is a simple solution to this problem, but it seems to me that this would be rare issue.

pkerpedjiev:
That's fair. Another way to do this would be to get the return values from each track's click handler synchronously and then create a dialog from the returned values.

I actually implemented something like that in the resgen-flavored higlass a while ago. Here's the code:

https://github.com/reservoirgenomics/resglass/blob/main/app/scripts/HiGlassComponent.js#L4553

Each track has a synchronous `click` handler that returns some data. The contents of the `app.click` event get populated with those return values. Because the tracks return data synchronously rather than subscribing to the `app.click` event, an outside caller (i.e. application) can subscribe and get notified of click events and then process data from them. Resgen uses this for annotations so that when you click on an annotation, the resgen viewer will show its details in the sidebar.

In this case, you could have tracks implement a `clickDialog` function that would return a React component. In `HiGlassComponent` you could call that function alongside the track's `click` function and build up a list of components to display. That would make it possible to display values from multiple overlaid tracks simultaneously in one dialog.

What do you think?

If you're OK with this approach, I could create a PR for the click handler that's in resglass and then you could stack a PR on top of that with the dialog code?

I think the dialog code is super useful, by the way, we need something like that for the pileup plugin. Event handling is tricky though and I'd prefer we did it in a way that's robust and can handle cases that don't happen often (like having multiple tracks wish to display some dialog data).

pkerpedjiev:
Great! I'll try to get a PR up this weekend.

pkerpedjiev:
Here it is: https://github.com/higlass/higlass/pull/1098

It's still missing tests but it should provide you something to work off of.

PR #1095, opened 2022-04-15
chore: bump pub-sub-es to latest

manzt:
## Description

> Why is it necessary?

I've been working on several applications with both HiGlass/Gosling where I'm generating HTML on the fly for the web-browser. I'd like to avoid writing the HTML to disk, and instead load the content via a data url (e.g. `data:text/html,<doctype!><html><h1>Hey</h1></html>`). Currently this throws an unrecoverable error due to an edge case with the version of `pub-sub-es` that higlass is on. This can additionally be useful with `puppeteer` for taking screenshots:

```javascript
import puppeteer from "puppeteer";

let viewconf = JSON.stringify(...);
let html = `\
<!doctype html>
<head>
  <link rel="stylesheet" href="https://unpkg.com/higlass@1.7/dist/hglib.css">
  <script src="https://unpkg.com/higlass@1.7/dist/hglib.min.js"></script>
</head>
<body></body>
<script>
	hglib.viewer(document.body, JSON.parse(\`${viewconf}\`);
</script>
</html>`;

let browser = await puppeteer.launch();
let page = await browser.newPage();
await page.setContent(html, { waitUntil: "networkidle0" });
let component = await page.waitForSelector(".higlass");
await component.screenshot({ path: "./example.jpeg" });
await browser.close();
```

> Details

`pub-sub-es` instantiates a browser `BroadcastChannel` if available _on import_ to broadcast events from the `gloablPubSub`. Instantiating a `BroadcastChannel` throws if the website does not contain an origin, which is the case when the page content _is_ the url. The previous version of `pub-sub-es` doesn't catch this edge case,

####  **pub-sub-es v1.2**
```javascript
const bc = (() => {
  const BC = window.BroadcastChannel;
  if (!BC) {
    console.warn("The Broadcast Channel API is not available in your browser.");
    return { postMessage: () => {} };
  }
  return new BC("pub-sub-es"); // throws because no origin, not caught
})();
```

while the latest version wraps the instantiation in a blanket `try`/`catch`.

#### **pub-sub-es v2**

```javascript
const bc = (() => {
  try {
    return new window.BroadcastChannel('pub-sub-es');
  } catch (e) {
    return { postMessage: () => {} };
  }
})();
```

### Breaking changes?

From what I can tell, the breaking changes in v2 are changes to options parameter for `createPubSub`, which higlass does not rely on. @flekschas 

https://github.com/flekschas/pub-sub/blob/master/CHANGELOG.md

## Checklist

- [ ] Updated CHANGELOG.md

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2399,7 +2399,8 @@
     "balanced-match": {
       "version": "1.0.0",
       "resolved": "https://registry.npmjs.org/balanced-match/-/balanced-match-1.0.0.tgz",
-      "integrity": "sha1-ibTRmasr7kneFk6gK4nORi1xt2c="
+      "integrity": "sha1-ibTRmasr7kneFk6gK4nORi1xt2c=",
+      "dev": true
     },
     "base": {
       "version": "0.11.2",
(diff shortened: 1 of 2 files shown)


flekschas (approved):
Should be a safe update 👍

manzt:
Any chance we could cut a release? @pkerpedjiev

PR #1102, opened 2022-06-25
Use webpack 5

pkerpedjiev:
## Description

Gargantuan PR with a bunch of webpack and package updates. Long overdue.

- Upgrade to webpack 5
- Upgrade a bunch of other libraries

> Why is it necessary?

- Getting with the times

Fixes #___

## Checklist

- [x] Updated CHANGELOG.md


test/utils/FetchMockHelper.js
```diff
@@ -1,5 +1,17 @@
 import fetchMock from 'fetch-mock';
 
+const timeout = (prom, time, ret) =>
```
pkerpedjiev, line 3: This is a failsafe to make sure that the FetchMockHelper doesn't time out itself.

app/scripts/HorizontalChromosomeLabels.js
```diff
@@ -451,6 +448,7 @@ class HorizontalChromosomeLabels extends PixiTrack {
 
       this.drawBoundsTicks(x1, x2);
 
+      console.log('tickTexts', this.tickTexts);
```
manzt, line 451: I'm guessing this is vestigial

```suggestion

```

package.json
```diff
@@ -37,44 +37,39 @@
...
+    "d3-transition": "^3.0.1",
+    "d3-zoom": "^3.0.0",
     "dom-scroll-into-view": "^1.2.1",
     "enzyme-adapter-react-16": "^1.2.0",
     "eventemitter3": "^2.0.2",
     "genbank-parser": "^1.0.0",
     "ndarray": "^1.0.18",
     "pako": "^1.0.10",
     "path": "^0.12.7",
+    "postcss": "^8.4.6",
```
manzt, line 66: probably want this as dev-dependency.

test/AxisTests.js
```diff
@@ -19,19 +19,22 @@ import {
 import viewconf from './view-configs/axis';
 
 configure({ adapter: new Adapter() });
+jasmine.DEFAULT_TIMEOUT_INTERVAL = 30000;
```
manzt, line 22: Is this something that should be set per test or globally?

test/AxisTests.js
```diff
@@ -19,19 +19,22 @@ import {
...
-describe('Simple HiGlassComponent', () => {
+describe('Axis tests', () => {
   let hgc = null;
   let div = null;
 
   const fetchMockHelper = new FetchMockHelper(null, 'AxisTests');
 
-  describe('Axis texts', () => {
-    beforeAll(async (done) => {
+  describe('Axis tets', () => {
```
manzt, line 30:
```suggestion
  describe('Axis tests', () => {
```

package.json
```diff
@@ -85,86 +80,90 @@
...
+    "eslint": "^8.7.0",
+    "eslint-config-airbnb": "^19.0.4",
+    "eslint-config-prettier": "^8.3.0",
     "eslint-plugin-import": "^2.14.0",
     "eslint-plugin-jsx-a11y": "^6.1.2",
-    "eslint-plugin-prettier": "^3.1.2",
-    "eslint-plugin-react": "7.11.1",
+    "eslint-plugin-prettier": "^4.0.0",
+    "eslint-plugin-react": "^7.14.2",
     "fast-css-loader": "^1.0.2",
```
manzt, line 129: It looks like `fast-css-loader` was removed in the webpack config. Can we delete it?

webpack.config.js
```diff
@@ -176,43 +188,26 @@ module.exports = (env, argv) => ({
...
-      amd: 'react',
-      root: 'React',
-    },
-    'react-dom': {
-      commonjs: 'react-dom',
-      commonjs2: 'react-dom',
-      amd: 'react-dom',
-      root: 'ReactDOM',
-    },
-  },
```
manzt, line 198: I'd recommend keeping these dependencies external – although I know it seems to cause headaches. Ideally, `higlass` should bundle it's dependencies, but since the export is a UMD module it seems necessary.

In Gosling, we use a different version of react/react-dom and PIXI (but we rely on accessing `PIXI` from `HGC.libraries.PIXI`). Changing this behavior means we would no longer specify the version of PIXI we are using. 

Additionally (and the bigger concern) is that we extend the `HiGlassComponent` directly in `gosling.js`, and render our own components with `react`/`react-dom`. From [the docs](https://reactjs.org/warnings/invalid-hook-call-warning.html#duplicate-react),

> In order for Hooks to work, the `react` import from your application code needs to resolve to the same module as the `react` import from inside the `react-dom` package.

Not externalizing `react` & `react-dom` leads to multiple copies React that we can't resolve since the code is bundled.

manzt (commented):
Thanks for all your work on this! Tried to give close review but there are lots of moving parts.
My main concern is the removal of `peerDependencies` and `external` in the webpack config. I think that removing react as a peerDependency will break projects that `npm i higlass` and use the `HiGlassComponent` directly.

Otherwise the changes look good to me (really happy about the d3 updates!), but I couldn't seem to get `npm run compile` to work.

pkerpedjiev:
To start, I think your overall sentiment that this change shouldn't be included in this PR is correct. It should be in its own PR to warrant proper discussion and easy rollback. I've already removed those changes from this PR. But since we've started the conversation, let's keep it going...

> Tried to give close review but there are lots of moving parts. My main concern is the removal of peerDependencies and external in the webpack config. I think that removing react as a peerDependency will break projects that npm i higlass and use the HiGlassComponent directly.

Why would it break projects that use the HiGlassComponent directly? You mention this line of reasoning in an inline comment:

> Not externalizing react & react-dom leads to multiple copies React that we can't resolve since the code is bundled.

If I recall correctly, is actually caused by dependencies including different version of React. `[react-draggable](https://github.com/react-grid-layout/react-draggable/blob/master/package.json#L73), for example, includes `React` as a dev dependency. As long as all components use the same version of react, it should be shaken out by the bundler. I could definitely be wrong here though. The reason I was leaning toward removing peer deps is because I was running into this exact same error.

The ultimate solution will be to update higlass to use the latest version of React. I've [started a PR](https://github.com/higlass/higlass/pull/1104) to get that ball rolling but it will need some work.

manzt:
Just to be clear, making a dependency external in the webpack bundle or a peer dependency are two different things. We could make `react`/`react-dom` and `pixi.js` regular dependencies but still keep external in webpack and probably not run into issues I highlighted. 

In fact, everything essentially a _development dependency_ in `higlass` except for `react/`react-dom`/`pixi.js` due to the way it is packaged as a webpack bundle. When you `npm install higlass`, all `dependencies` in the `package.json` are installed in `node_modules`, but every single one of these dependencies are duplicated already in the `dist/hglib.js`. The effect is that the developer using `higlass` _only_ has control over the dependencies which marked `external` in the webpack config. E.g., a patch version of `d3-selection` is released with a security fix? Need to wait for a new version of `higlass` to be released that bumps the version; you cannot upgrade on your own. This is the primary issue with bundling dependencies.

In the case of `higlass`, since `react`/`react-dom`/`pixi.js` are marked as external end users have control over the versions. It just so happens these dependencies also happen to be `peerDependencies` in the `package.json`, and therefore must be manually installed in the repo (at least historically – as of npm v7 peerDeps are installed automatically if missing.)

> Why would it break projects that use the HiGlassComponent directly?

The issue is that in order for react hooks to work, there must only be one react/react-dom for the component tree. https://reactjs.org/warnings/invalid-hook-call-warning.html#duplicate-react. Even if there are multiple reacts but they are the same exact version, hooks will not work.

> [react-draggable](https://github.com/react-grid-layout/react-draggable/blob/master/package.json#L73), for example, includes React as a dev dependency. 

This is true, but the distribution of `react-draggable` has `react` and `react-dom` [as external in the final bundle](https://unpkg.com/react-draggable@4.4.5/build/web/react-draggable.min.js):

```diff
function webpackUniversalModuleDefinition(root, factory) {
	if(typeof exports === 'object' && typeof module === 'object')
+		module.exports = factory(require("react"), require("react-dom"));
	else if(typeof define === 'function' && define.amd)
+		define(["react", "react-dom"], factory);
	else if(typeof exports === 'object')
+		exports["ReactDraggable"] = factory(require("react"), require("react-dom"));
	else
+		root["ReactDraggable"] = factory(root["React"], root["ReactDOM"]);
})(self, function(__WEBPACK_EXTERNAL_MODULE__359__, __WEBPACK_EXTERNAL_MODULE__318__) {
```

The effect of specifying `react`/`react-dom` as dev-dependencies but keeping external in the published module is that users must (implicitly) install react/react-dom themselves (e.g,. there will be no warning if you `npm install react-draggable` and don't have react/react-dom). Peer-dependencies allows you to say "you need to have this installed" but also specifics which versions are compatible: `"react": "^16.0.0 | ^17.0.0 | ^18.0.0"`.

> As long as all components use the same version of react, it should be shaken out by the bundler. I could definitely be wrong here though.

This sadly is not true. The moment that you bundle a dependency in a library, a bundler cannot treeshake it or resolve to shared dependencies. Bundler rely on static analysis of imports, which is easier to show with ESM than UMD modules:

#### Example `node_modules`

```javascript
// foo.js
export * as React from "react"; // "external"
```

```javascript
// bar.js
export * as React from "react"; "external"
```

```javascript
// baz.js
export const React = { /* ... */ }; // *not* "external", React is bundled in baz.js
```

#### Source code

```javascript
// index.js
export { React as React0 } from "foo";
export { React as React1 } from "bar";
export { React as React2 } from "baz";
```

#### Final bundle

```javascript
// bundle.js
const React_external = { /* ... */ };
export const React0 = React_external; // resolve to same react
export const React1 = React_external; // resolve to same react
export const React2 = { /* ... * }; // different react because it was bundled in the library!
```

pkerpedjiev:
Thanks for the detailed explanation. It makes sense that dependencies are not actually tree shaken. If a dependency bundle is minified, how would the bundler know which code to dedup?

Regarding external / peer dependencies, that also makes sense. I wish I had taken notes on how to resolve the multiple react issue when I encountered it a few months ago. To the best of my memory, it involved adding "resolve" sections to my webpack file.

Another annoyance with external dependencies is that if you have an application that looks like this:

```
  <link rel="stylesheet" href="https://unpkg.com/higlass@1.11/dist/hglib.css">
  <script src="https://unpkg.com/higlass@1.11/dist/hglib.js"></script>
```

And we decide to bump the version of higlass to use a later version of React, then you, as the application developer, need to know to not only update the higlass version in your application but the versions of the react imports. So essentially updating the react dependency version requires not only updating the package.json, but the docs and any other application that has it included.

For the purposes of this PR and the near future I'm fine with leaving them as external.

manzt:
> If a dependency bundle is minified, how would the bundler know which code to dedup?

Even unminified, bundlers aren't smart enough to recognize that code is duplicated from a module. Bundlers essentially rely totally on static analysis of imports/exports for dedupe. For example if one dependency includes,

```javascript
import { foo } from "bar";
```

Than anywhere your source code or other dependencies import from `"bar"` can be deduplicated in the resulting bundle (if their `package.json` specify compatible semvar of `bar`). That's it. You can think of the current `higlass` package as:

```javascript
// dist/hglib.js
import React from "react";
import ReactDOM from "react-dom";
import * as PIXI from "pixi.js";

/* all of higlass and other dependencies */

export { HiGlassComponent, viewer, /* ... */ }
```

so _only_ `react`, `react-dom`, and `pixi.js` are eligible to be de-duplicated.

> To the best of my memory, it involved adding "resolve" sections to my webpack file.

Yes, this is common. If there are multiple packages which list `react` as a dependency (rather than a `peerDependency`), each will install a separate version of `react`. This is desirable behavior generally because dependencies could rely on different semvers of the shared dep which are incompatible.

The `resolve` field in the webpack config can tell the bundler to resolve any import from `"react"` to a specific directory in `node_modules`. This is you telling the bundler "I know this version of `react` is ok for all my dependencies". However, if `react` wasn't external for `higlass`, the `resolve` doesn't do anything (there is no `"react"` string to resolve; all the code is inlined).

> Another annoyance with external dependencies is that if you have an application that looks like this,

I agree this is annoying, but only if you need to use `higlass` without a bundler. This was the old school way of packaging things for the browser and led to module systems like `amd`. It's becoming more an more common to ship packages as ESM (preserving `import React from "react"` in what is published to npm) and then CDNs now bundle/minify the code.

The real issue is that there are two use cases for `higlass`:

1.) A simple HTML file with a script tag that gives you access to `hglib.viewer`.
2.) An application where you have installed `higlass` as a dependency and are bundling it yourself.

Ideally we would ship a module for the latter (keeping all `import/exports`) and a fully bundled/minified UMD script for the former. Odds are if you are using `higlass` through a script tag, you aren't using something like `HiGlassComponent` directly but a top-level utility like `hglib.viewer` (which owns the entire react render tree). In that senario it would be totally ok to bundle react/react-dom/pixi.js, it's just doing so hurts the use of `higlass` with a bundler.

test/AxisTests.js
pkerpedjiev: Good idea. I'll set it globally.

package.json
pkerpedjiev: yup

app/index.html
```diff
@@ -3,7 +3,6 @@
 <meta charset="utf-8">
 <meta name="viewport" content="width=device-width,initial-scale=1">
 <title>HiGlass</title>
-<link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css">
```
flekschas, line 6: Was bootstrap finally removed? If so might be worth mentioning in the changelog

app/scripts/HorizontalChromosomeLabels.js
flekschas: Why is log message being added?

package.json
```diff
@@ -37,30 +37,25 @@
   "engines": {
     "node": ">=0.12.0"
   },
-  "peerDependencies": {
-    "pixi.js": "^5.0.3",
-    "react": "^16.6.3",
-    "react-bootstrap": "0.32.1",
-    "react-dom": "^16.6.3"
-  },
```
flekschas, line 45: I still don't understand why we're force bundling `react` and `react-dom` now. Maybe I am missing something but whenever you search for the purpose of `peerDependencies`, one finds `react` and `react-dom` [as](https://classic.yarnpkg.com/lang/en/docs/dependency-types/#toc-peerdependencies) [the](https://code-trotter.com/web/dependencies-vs-devdependencies-vs-peerdependencies/) [examples](https://stackoverflow.com/questions/30451556/what-is-the-correct-way-of-adding-a-dependency-to-react-in-your-package-json-for). If someone else wants to use HiGlass in their own React app they'll potentially end up with two copies of React in their final bundle. Also, as of npm v7 peer dependencies are installed automatically so there isn't even a convenience difference.

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Release notes
 
+## Unreleased
+
+- Update to webpack5
+- Update a bunch of packages (most notably d3 packages which have a slightly different API)
+- Fix tests
+- Update scss to fix deprecation warnings
+- Remove pixi.js, react and react-dom as external packages
```
flekschas, line 9: I am confused here. 9 days ago or so, externals were added back to `webpack.config.js`. Is this an outdated bullet point?

package.json
flekschas: Maybe it doesn't matter as the packages are defined as externals in webpack. I guess I would just find it cleaner if they were marked as `peerDependencies` as well but whatever.

flekschas (commented):
Left some clarification comments

package.json
pkerpedjiev: Alright. Added back.

CHANGELOG.md
pkerpedjiev: No, just didn't update the CHANGELOG to reflect that change. Good catch!

app/index.html
pkerpedjiev: Yeah, it's gone. It's been gone for a while but this stylesheet link wasn't removed.

I'm not going to add it to the CHANGELOG because it would imply that bootstrap was removed in this PR when it wasn't.

manzt:
Any way I can help push this forward?

pkerpedjiev:
Superseded by https://github.com/higlass/higlass/pull/1111

pkerpedjiev:
Yes, it’s ready to be merged. I just didn’t do it because the tests were
failing but now they’re not so feel free to merge.

PR #1111, opened 2022-08-26
refactor: migrate webpack to vite, karma to web-test-runner

manzt:
## Description

Be gone webpack and karma! This PR extends on top of #1102, but massively simplifies the current testing/build suite for HiGlass. 

### Primary changes 

- Webpack 5 is replaced with [Vite](https://vitejs.dev/). `npm start` is an alias for `vite`
- Karma (and Jasmine) are replaced with `@web/test-runner` (mocha test runner) with `chai` for assertions.

### Secondary

- fix use of `d3` to reflect latest (upgraded API).  Several parts of the code base relied on `event` from `d3-select` which is no longer supported. I migrated those modules to use the latest API.
- files with JSX now have `.jsx` file extensions. This is required for vite.

> Why is it necessary?

Removing lots of painful dev dependencies. Much more simple to maintain and upgrade (no timeouts with karma/jasmine). Faster dev, faster tests, faster build.

## TODO:

- [ ] fix remaining tests
- [x] update "compile" script
- [x] remove `webpack.*` configs and `karma.conf` are the AMD configs used at all???

## Checklist

- [ ] Updated CHANGELOG.md


app/scripts/CenterTrack.jsx
```diff
@@ -27,15 +32,15 @@ class CenterTrack extends React.Component {
...
+    this.brushBehaviorX = brushX()
       .on('brush', this.brushedX.bind(this))
       .on('end', this.brushedXEnded.bind(this));
 
-    this.brushBehaviorY = brushY(true, true, true)
+    this.brushBehaviorY = brushY()
       .on('brush', this.brushedY.bind(this))
       .on('end', this.brushedYEnded.bind(this));
 
-    this.brushBehaviorXY = brush(true)
```
manzt, line 38: This boolean API comes from Fritz's d3-brush fork which we no longer rely on.

app/scripts/CenterTrack.jsx
```diff
@@ -170,13 +170,13 @@ class CenterTrack extends React.Component {
...
-  brushedX() {
+  brushedX(event) {
     // Need to reassign variable to check after reset
     const rangeSelectionMoved = this.rangeSelectionMoved;
     this.rangeSelectionMoved = false;
 
     if (
-      !this.sourceEvent ||
+      !sourceEvent(event) ||
       !this.props.onRangeSelectionX ||
```
manzt, line 180: https://observablehq.com/@d3/d3v6-migration-guide

manzt: hmm I'm a bit confused on this... `package.json` has older versions of d3 vite seems to be resolving to some later version. It would probably (actually be easier) if we merged your webpack changes first, and then rebased this PR on that.

manzt:
Ok I'm getting very close to having all the tests working. Just four failing and I've re-enabled all the tests).

image


The issue withe `Plugin*Tests.js` is that the assertion library cannot convert a JS `Module` into plain objects to compare between `dataFetcher.hgc` and `trackRenderer.availableForPlugins`.

manzt:
Fixed `APITests.js` by setting a longer max duration for each test file. We could probably break that file up into more to improve speed.

app/scripts/CenterTrack.jsx
manzt: https://github.com/d3/d3-selection/releases/tag/v2.0.0

package.json
```diff
@@ -17,17 +17,12 @@
...
-    "build": "rm -rf build/*; rm -rf dist/*; mkdir -p dist/; webpack --config ./webpack.config.js --mode development; cp build/hglib.js dist/; cp app/index.html dist/; cp build/hglib.css dist/hglib.css;",
-    "build-amd": "rm -rf build/*; rm -rf dist/*; mkdir -p dist/; webpack --config ./webpack.config.amd.js --mode development; cp build/hglib-amd.js dist/; cp build/hglib-amd.css dist/hglib-amd.css;",
-    "build-amd-dev": "rm -rf build/*; rm -rf dist/*; mkdir -p dist/; webpack --config ./webpack.config.amd.dev.js --mode development; cp build/hglib-amd-dev.js dist/; cp build/hglib-amd-dev.css dist/hglib-amd-dev.css;",
-    "build-watch": "rm -rf build/*; rm -rf dist/*; mkdir -p dist/; webpack --config ./webpack.config.js --mode development --watch; cp build/hglib.js dist/; cp app/index.html dist/; cp build/hglib.css dist/hglib.css;",
+    "test": "wtr",
+    "start": "vite",
+    "build": "vite build",
     "compile": "rm -rf build/*; rm -rf dist/*; mkdir -p dist/; webpack --config ./webpack.config.js --mode production; cp build/hglib.min.js dist/; cp build/hglib.js dist/; cp app/index.html dist/; cp build/hglib.css dist/hglib.css; zip -r dist.zip dist",
```
pkerpedjiev, line 23: Should this be changed or removed? Looks like it still uses webpack.

manzt: Fixed. I wrote `./scripts/compile.mjs` to minify the js output and generate an `index.html` which uses the UMD bundle.

pkerpedjiev:
This PR is huge but we discussed on Zoom. I say merge the sucker and let's fix any issues in separate PRs.

PR #1114, opened 2022-08-30
chore: remove unused FetchMockHelper

manzt:
## Description

Removes the use of `FetchMockHelper.js` from our test suite as it no longer seems to be necessary.
In the future to maybe speed up tests by avoiding network requests, we could intercept
requests to `higlass.io/api/v1/*` with something like a [`ServiceWorker`](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API/Using_Service_Workers), within `web-test-runner`
and create a persistent cache to be reused across all our test suite, e.g.,

```javascript
const cacheFirst = async (request) => {
  const responseFromCache = await caches.match(request);
  if (responseFromCache) {
    return responseFromCache;
  }
  return fetch(request);
};

self.addEventListener("fetch", (event) => {
  event.respondWith(cacheFirst(event.request));
});
```

or add a route to Vite to handle some special requests from the client.

Removes `FetchMockHelper` and the `mockedFetchPlugin` from vite.

> Why is it necessary?

Fixes #___

## Checklist

- [ ] Updated CHANGELOG.md


pkerpedjiev:
Just for some background, it's not used because I commented it out to get the webpack5 upgrade running. It may work now with vite.

It's actually a nice feature to have. I'd be OK with a different implementation too.

manzt:
I see. I have an idea for an alternative implementation but just haven't gotten around to it yet. I think we should be able to crawl the tests, intercept requests to `higlass.io` and store the JSON responses to disk (perhaps even check that file into the repo). Then we can proxy requests to higlass.io from the client and use this file as a persistent cache.

pkerpedjiev:
> perhaps even check that file into the repo

Yeah, that would be key. Ideally one would run the tests locally to fetch the data. We would store that data to disk and check it in. Then when the tests are run on CI, they wouldn't need to hit an external service. Bonus points if we can set the tests to only use the local cache when running in CI. That would both ensure that we've fetched all the data and make sure that the tests run in a timely manner without waiting for an external service for data.

manzt:
Have a look at #1118, I have implemented this... but the requests do not seem to be deterministic from the test suite. Every time I run the tests, the cache expands in size, and cache-hit rate ~50% for a freshly created cache.

Are there any query params besides the tile id (e.g., `s`?) which might make caching responses using the client URL (as the cache key) an issue?

pkerpedjiev:
Just to provide a bit of context, the `s=` query string parameter was meant to identify the session but we never actually use it for anything. That's the only parameter that should change between requests. The only other possibility is that if your window is varying in size then the calculated zoom level may be different between runs.

If possible, could you isolate the cache misses to one test and see which ids get added?

PR #1115, opened 2022-08-30
chore: move test utilities outside of hglib

manzt:
## Description

The exported `test-helpers.js` from `app/scripts/utils` includes
"enzyme" as a dependency, which adds about ~1Mb (0.5 minified) to the 
3.7Mb (1.9 minified) totoal of HiGlass.

If this is is only used internally, I think we should move these methods to 
the `/test/` folder and outside the final bundle.

#### Before

```
higlass on  develop [$!?] is 📦 v1.11.11 via  v18.1.0 took 11s

❯ ls -la dist
.rw-r--r--  51k manzt 30 Aug 15:35 hglib.css
.rw-r--r-- 3.7M manzt 30 Aug 15:35 hglib.js
.rw-r--r-- 1.9M manzt 30 Aug 15:35 hglib.min.js
.rw-r--r-- 8.8k manzt 30 Aug 15:35 index.html
```

image

#### After

```
higlass on  develop [$✘!?] is 📦 v1.11.11 via  v18.1.0 took 13s

❯ ls -la dist
.rw-r--r--  51k manzt 30 Aug 15:58 hglib.css
.rw-r--r-- 2.7M manzt 30 Aug 15:58 hglib.js
.rw-r--r-- 1.4M manzt 30 Aug 15:58 hglib.min.js
.rw-r--r-- 8.8k manzt 30 Aug 15:58 index.html
```

image

Moved `app/scripts/utils/tests-helpers.jsx` to `test/utils/test-helpers.jsx` and 
remove exports from `hglib.js`.

> Why is it necessary?

Fixes #___

## Checklist

- [ ] Updated CHANGELOG.md


pkerpedjiev:
Would it be possible to export those functions that you removed from hglib as another library (hglib.test.js or something to that effect)? They can be useful to plugin libraries that wish to implement tests. I don't know which ones actually do so off the top of my head.

Another potential solution: pass in the `mount` function as an argument to remove the `enzyme` dependency.

scripts/compile.mjs
```diff
@@ -56,3 +56,12 @@ await fs.promises.writeFile(
     <script src="./hglib.min.js"></script>
   `)
 );
+
+// =========================================
+//       Copy test helpers to dist
+// =========================================
+
+await fs.promises.copyFile(
```
manzt, line 64: We just copy this file to dist. We shouldn't need to bundle since `higlass` is already bundled by us

package.json
```diff
@@ -25,6 +26,10 @@
     "lint": "eslint app test vite.config.js web-test-runner.config.mjs",
     "precommit": "NODE_ENV=production lint-staged"
   },
+  "exports": {
+    ".": "./dist/hglib.js",
+    "./testing": "./dist/testing.js"
+  },
```
manzt, line 32: this is a package entrypoint. Consumers of `higlass` will be able to 

```js
import * as testingUtils from 'higlass/testing';


```

manzt: https://nodejs.org/api/packages.html#package-entry-points

test/utils/test-helpers.js
```diff
@@ -213,8 +215,8 @@ export const mountHGComponent = (prevDiv, prevHgc, viewConf, done, options = {})
   div.setAttribute('style', style);
   div.setAttribute('id', 'simple-hg-component');
 
-  const hgc = mount(
-    <HiGlassComponent options={{ bounded }} viewConfig={viewConf} />,
+  const hgc = Enzyme.mount(
+    React.createElement(HiGlassComponent, { options: { bounded }, viewConfig }),
```
manzt, line 219: manually create element so consumers don't need to configure JSX transform.

test/utils/test-helpers.js
```diff
@@ -1,20 +1,16 @@
...
-  getTrackObjectFromHGC,
+  requestsInFlight,
   getTrackRenderer,
-} from './get-higlass-components';
-
-import HiGlassComponent from '../HiGlassComponent';
+  getTrackObjectFromHGC,
+  HiGlassComponent,
+// eslint-disable-next-line import/no-unresolved
+} from 'higlass';
```
manzt, line 13: This file will be available to consumers in the final build under `dist/testing.js`. Users should be able to import with. 

During development, Vite aliases `higlass` to `app/scripts/hglib.jsx` but in the final build consumers will get `dist/hglib.js`.

manzt:
> Would it be possible to export those functions that you removed from hglib as another library (hglib.test.js or something to that effect)? They can be useful to plugin libraries that wish to implement tests. I don't know which ones actually do so off the top of my head.

Yes. I have now implemented this by adding a `./testing` entrypoint to the `higlass` package (`./dist/hglib.test.js`).

manzt:
Another option would be to move `test/utils/test-helpers.js` to the root of the repo and include this file in `package.json` files. That way consumers would have:

```js
import * as hglib from 'higlass';
import * as helpers from 'higlass/test-helpers.js';
```

pkerpedjiev:
Of the two options you mentioned, which one do you think is better? They seem more or less equivalent to me so the only thing I'd judge on is aesthetics and in that regard the second option wins.

test/HorizontalHeatmapTests.js
```diff
@@ -1,21 +1,13 @@
 /* eslint-env mocha */
-import {
-  configure,
-  // render,
-} from 'enzyme';
-
+import Enzyme from 'enzyme';
```
pkerpedjiev, line 2: Just curious, why this change?

scripts/compile.mjs
```diff
@@ -56,3 +56,12 @@ await fs.promises.writeFile(
     <script src="./hglib.min.js"></script>
   `)
 );
+
```
pkerpedjiev, line 59: What is this file?

pkerpedjiev (approved):
👍 Just a couple questions

scripts/compile.mjs
manzt: This file is executed within `npm run compile`. First we build the library with Vite, and then this script adapts those outputs to create identical outputs to the previous from webpack. There is probably a way to do this all with a complicated Vite config, but I think this is much easier to read/understand.

test/HorizontalHeatmapTests.js
manzt: No reason, really. I just found it easier to figure out where the setup/teardown code came from.

manzt:
Closing because of all the changes with our testing. I think this would still be great but can follow up in a separate PR.

PR #1118, opened 2022-09-02
[WIP] cache client-side responses with plugin

manzt:
## Description

Replaces `FetchMockData` with a plugin that sniffs for requests for higlass.io/resgen.io and caches responses. The plugin works in two parts:

1.) a client script which replaces global `fetch` with a `fetch` that redirects to `localhost:8000/@cache` if the request URL is for higlass.io/resgen.io.
2.) a server-side middleware that looks for requests for `/@cache` and checks a persistent cache for the response from that URL. The client has the option to make a POST request to this endpoint to add data to the cache.

On teardown, the in-memory cache is written to disk (`response-cache.json`) and read into memory before each test. 

This is a WIP because the cache-hit rate (after initial run) seems to be <55% which seems odd. I'd think that request URLs should be consistent across tests, but I'm not familiar with HiGlass server enough. 

EDIT: every time I run the test `cache-response.json` get larger, so almost certainly there is part of the higlass/resgen URL (which is used as the cache key) which varies between test runs. Writing a `cachKey(url)` in the plugin would allow us to reuse reponses (if appropriate) for multiple requests.

- added `scripts/cache-plugin.mjs`

> Why is it necessary?

Fixes #___

## Checklist

- [ ] Updated CHANGELOG.md


manzt:
I'm realizing this is basically a local CDN. Does something like that already exist for testing that we could proxy requests through and store?

pkerpedjiev:
No idea. But who cares if it's basically a local CDN. If it works let's use it!

PR #1120, opened 2022-09-07
fix(test): HGC comparison tests for plugins

manzt:
## Description

Fixes broken tests in `PluginTrackTests.js` and `PluginDataFetcherTests.js`.

I'm not actually sure how this test was passing previously... The tests both check to make sure that `trackRenderer.avaiableForPlugins` is the same as `plugin.hgc`, BUT the `TrackRenderer` mutates HGC to add additional servies at runtime

- (prior to #1111) https://github.com/higlass/higlass/blob/04df604872ea820a190995f2b34f92fcd552a68b/app/scripts/TrackRenderer.js#L119-L121
- (after #1111) https://github.com/higlass/higlass/blob/dc2098eb4e738903fbc7da2dd0589416dced80d5/app/scripts/TrackRenderer.jsx#L119-L126

so `AVAILABLE_FOR_PLUGINS` will never be the same as `trackRenderer.avaiableForPlugins`.

use `trackRenderer.avaiableForPlugins` when initiating plugins instead of global `AVAILABLE_FOR_PLUGINS`.

> Why is it necessary?

Fixes 2/4 remaining failing tests.

## Checklist

- [ ] Updated CHANGELOG.md

### DIFF
--- a/app/scripts/TrackRenderer.jsx
+++ b/app/scripts/TrackRenderer.jsx
@@ -1504,6 +1504,7 @@ class TrackRenderer extends React.Component {
       dataConfig,
       this.props.pubSub,
       this.props.pluginDataFetchers,
+      this.availableForPlugins,
     );
 
     // To simplify the context creation via ES6 object shortcuts.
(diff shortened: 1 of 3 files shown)


pkerpedjiev:
I think this is fine. There was some reason for this test but I don't remember what it was. We'll find out if/when it bites us.

pkerpedjiev (approved):

PR #1123, opened 2022-09-07
fix(test): Update final x position of exported SVG

manzt:
## Description

Fixes the broken SVG export test. 

The expect x/y positions for the expected svg elements are slightly different. I'm guessing this is due to the new testing environment and not a regression. I updated the positions in the test from this correct SVG export:

image

> Why is it necessary?

Fixes the last outstanding broken test.

Fixes #___

## Checklist

- [ ] Updated CHANGELOG.md

### DIFF
--- a/test/HiGlassComponent/divergent-track.js
+++ b/test/HiGlassComponent/divergent-track.js
@@ -34,10 +34,10 @@ describe('Divergent tracks', () => {
   it('Check that there are green and red rects', async () => {
     const svg = hgc.instance().createSVG();
     expect(
-      svg.querySelector("rect[fill='green'][stroke='green'][x='11.249637595676889']")
+      svg.querySelector("rect[fill='green'][stroke='green'][x^='11.24963759567']")
     ).to.exist;
     expect(
-      svg.querySelector("rect[fill='red'][stroke='red'][x='29.818754489547988']")
+      svg.querySelector("rect[fill='red'][stroke='red'][x^='29.81875448954']")
     ).to.exist;
   });
 });
--- a/test/HiGlassComponent/divergent-track.js
+++ b/test/HiGlassComponent/divergent-track.js
@@ -3,7 +3,7 @@ import Enzyme from 'enzyme';
 import Adapter from '@wojtekmaj/enzyme-adapter-react-17';
 import { expect } from 'chai';
 
-import { mountHGComponent, removeHGComponent } from '../../app/scripts/utils';
+import { mountHGComponentAsync, removeHGComponent } from '../../app/scripts/utils';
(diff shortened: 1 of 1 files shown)


test/HiGlassComponent/divergent-track.js
```diff
@@ -31,17 +31,13 @@ describe('Divergent tracks', () => {
...
 
-  it('Check that there are green and red rects', (done) => {
+  it('Check that there are green and red rects', async () => {
     const svg = hgc.instance().createSVG();
-
-    const svgText = new XMLSerializer().serializeToString(svg);
     expect(
-      svgText.indexOf('fill="green" stroke="green" x="11.24963759567723"'),
-    ).to.be.greaterThan(0);
+      svg.querySelector("rect[fill='green'][stroke='green'][x='11.249637595676889']")
```
manzt, line 37:
```diff
11.24963759567723
11.249637595676889
```

test/HiGlassComponent/divergent-track.js
```diff
@@ -31,17 +31,13 @@ describe('Divergent tracks', () => {
...
+    ).to.exist;
     expect(
-      svgText.indexOf('fill="red" stroke="red" x="29.818754489548308"'),
-    ).to.be.greaterThan(0);
-
-    done();
+      svg.querySelector("rect[fill='red'][stroke='red'][x='29.818754489547988']")
```
manzt, line 40:
```
29.818754489548308 # prev
29.818754489547988 # now
```

test/HiGlassComponent/divergent-track.js
```diff
@@ -31,17 +31,13 @@ describe('Divergent tracks', () => {
...
 
-  it('Check that there are green and red rects', (done) => {
+  it('Check that there are green and red rects', async () => {
     const svg = hgc.instance().createSVG();
-
-    const svgText = new XMLSerializer().serializeToString(svg);
     expect(
-      svgText.indexOf('fill="green" stroke="green" x="11.24963759567723"'),
-    ).to.be.greaterThan(0);
+      svg.querySelector("rect[fill='green'][stroke='green'][x^='11.24963759567']")
```
manzt, line 37: the `x^=val` is a [substring match](https://www.w3.org/TR/selectors/#attribute-substrings) rather than exact match, allowing us to leave off some significant digits which were causing issues with the test passing.

test/HiGlassComponent/divergent-track.js
manzt: changed now with substring css selector https://www.w3.org/TR/selectors/#attribute-substrings

PR #1125, opened 2022-09-08
chore: remove react-css-modules

manzt:
## Description

**TL;DR - the ergonomics provided by `babel-plugin-react-css` are extremely limited and don't outweigh the long term development burden. Keeping it as a dependency means that we rely on babel (and it's compatibility with babel) for our build. Without it, we lose babel as a direct dependency and are more free to upgrade build tooling in the future.**

IMO this is an important PR to improve the long term maintenance of higlass. The [`babel-plugin-react-css`](https://github.com/gajus/babel-plugin-react-css-modules) isn't very well maintained and adds a JSX extension (`styleName`) which is largely unnecessary now with css module support in may build tools.

Additionally, the babel plugin mangles css names which must be unified with Vite's `css` output, so the `generateScopedName` requires additional configuration to be consistent between the two. Removing this plugin means that vite takes care of everything (and similarly would webpack or other modern build tool) without additional config.

- Removes the use of `styleName` from any JSX in favor of `className` using Vite's builtin css module support.
- Adds tiny [`clsx`](https://www.google.com/search?client=firefox-b-1-d&q=clsx) as a dependency for constructing `className` strings conditionally

```javascript
import clsx from 'clsx';

const className = clsx('foo', true && 'bar', { baz: true, buz: false });
// "foo bar baz"
``` 

with CSS modules, the use looks something like:

```javascript
import clsx from 'clsx';
import classes from './styles.module.scss';

const className = clsx('foo', {
  [classes.bar]: props.withBar,
  [classes['fiz-buz']]: props.withFizBuz,
});
```

> Why is it necessary?

Less config, no direct dependency on babel anymore.

Fixes #___

## Checklist

- [ ] Updated CHANGELOG.md

### DIFF
--- a/app/scripts/AddTrackPositionMenu.jsx
+++ b/app/scripts/AddTrackPositionMenu.jsx
@@ -4,56 +4,56 @@ import React from 'react';
 import withTheme from './hocs/with-theme';
 import { THEME_DARK } from './configs';
 
-import '../styles/AddTrackPositionMenu.module.scss';
+import classes from '../styles/AddTrackPositionMenu.module.scss';
 
 function AddTrackPositionMenu(props) {
-  let tableStyleNames = 'add-track-position-table';
+  const tableClassNames = [classes['add-track-position-table']];
   if (props.theme === THEME_DARK)
-    tableStyleNames += ' add-track-position-table-dark';
+    tableClassNames.push(classes['add-track-position-table-dark']);
   return (
     <div>
-      <div styleName="add-track-position-span">Add Track...</div>
-      <table styleName={tableStyleNames}>
+      <div className={classes["add-track-position-span"]}>Add Track...</div>
+      <table className={tableClassNames.join(" ")}>
         <tbody>
           <tr style={{ height: '30px' }}>
-            <td styleName="add-track-position-other" />
+            <td className={classes["add-track-position-other"]} />
             <td
+              className={classes["add-track-position-top-center"]}
               onClick={() => props.onTrackPositionChosen('top')}
-              styleName="add-track-position-top-center"
             >
               top
             </td>
-            <td styleName="add-track-position-other" />
+            <td className={classes["add-track-position-other"]} />
           </tr>
           <tr style={{ height: '80px' }}>
             <td
+              className={classes["add-track-position-middle-left"]}
               onClick={() => props.onTrackPositionChosen('left')}
-              styleName="add-track-position-middle-left"
             >
               left
             </td>
             <td
+              className={classes["add-track-position-middle-middle"]}
               onClick={() => props.onTrackPositionChosen('center')}
-              styleName="add-track-position-middle-middle"
             >
               center
             </td>
             <td
+              className={classes["add-track-position-middle-right"]}
               onClick={() => props.onTrackPositionChosen('right')}
-              styleName="add-track-position-middle-right"
             >
               right
             </td>
           </tr>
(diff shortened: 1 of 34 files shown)


app/scripts/AddTrackPositionMenu.jsx
```diff
@@ -4,56 +4,56 @@ import React from 'react';
...
+  const tableClassNames = [classes['add-track-position-table']];
   if (props.theme === THEME_DARK)
-    tableStyleNames += ' add-track-position-table-dark';
+    tableClassNames.push(classes['add-track-position-table-dark']);
   return (
     <div>
-      <div styleName="add-track-position-span">Add Track...</div>
-      <table styleName={tableStyleNames}>
+      <div className={classes["add-track-position-span"]}>Add Track...</div>
+      <table className={tableClassNames.join(" ")}>
```
manzt, line 16: We could add a tiny runtime dependency like [`classnames`](https://www.npmjs.com/package/classnames) to make this a little less clunky, e.g.,

```javascript
const className = classNames(classes['add-track-position-table'], {
	[classes['add-track-position-table-dark']]: props.theme === THEME_DARK,
})
```

app/scripts/Modal.jsx
```diff
@@ -5,28 +5,32 @@ import Button from './Button';
...
+    <div className={classNames.join(" ")}>
+      <div className={classes["modal-wrap"]}>
         <div
-          styleName={`modal-window ${
-            props.maxHeight ? 'modal-window-max-height' : ''
-          }`}
+          className={[
+            classes['modal-window'],
+            props.maxHeight ? classes['modal-window-max-height'] : ''
+          ].join(" ")}
```
manzt, line 26: here as well

```suggestion
          className={classNames(classes['modal-window'], {
            [classes['modal-window-max-height']]: props.maxHeight,
          })}
```

app/scripts/AddTrackPositionMenu.jsx
manzt: Update: I added the tiny [`clsx`](https://www.npmjs.com/package/clsx) as a dependency for this purpose.

app/scripts/TrackControl.jsx
```diff
@@ -1,39 +1,31 @@
...
-const getButtonClassNames = (props) => {
-  let buttonClassName = 'track-control-button';
-
-  buttonClassName += props.isVertical ? ' track-control-button-vertical' : '';
-
-  return buttonClassName;
-};
+const getButtonClassName = (props) => clsx("no-zoom", classes['track-control-button'], {
+  [classes['track-control-button-vertical']]: props.isVertical,
+});
```
manzt, line 23: Probably best example of use of `clsx`.

app/scripts/AddTrackPositionMenu.jsx
manzt: Update, I added [`clsx`](https://github.com/lukeed/clsx) instead.

app/scripts/TrackControl.jsx
pkerpedjiev: Love it!

pkerpedjiev:
Love it! Getting rid of babel makes me happy. The extra layer of indirection between the class names in the css and the react `className` prop is good and `clsx` is a nice touch.

Two very enthusiastic 👍👍

pkerpedjiev (approved):

PR #1127, opened 2022-09-10
chore: update formatting and linting

manzt:
## Description

Something to think about. 

This PR bootstraps a new eslint config for the repo via `npm init @eslint/config`, selecting react + airbnb defaults. There were a bunch of specific rules specified in the previous config (mostly react stuff) don't cause any errors. That in itself doesn't negate their value (they could have been enforcing more strict rules than the `plugin:react/recommended` defaults), but unless we feel very strongly about any of those rules my preference is for less config and better defaults. 

Just to clarify how eslint works:

- Plugins defines a set of linting rules
- `extends` allows you to opt-in to pre-configured sets of rules (recommended by plugins, etc) without manually needing to do so in `rules`
- `rules` trumps all and allows configuration of individual rules for your project

- `.eslintrc` -> `.eslintrc.js`
- Removed lots of react rules under `rules` in favor of those recommended by `eslint-plugin-react`

> Why is it necessary?

More simple config.

Fixes #___

## Checklist

- [ ] Updated CHANGELOG.md


manzt:
One other thing I realized is that the `eslint-prettier-plugin` does not actually run prettier. Instead it tells `eslint` to _ignore_ specific rules since prettier will take care of them. Unfortunately, I don't think that `prettier` has been run in some time and lots of files need to be prettified.

Our CI should run:

```
prettier --check 
eslint
```

and users should run 

```
prettier --write
eslint --fix
```

I'm happy to run this and update the workflow accordingly.

.github/workflows/ci.yml
```diff
@@ -35,6 +35,7 @@ jobs:
       - run: ajv validate -s app/schema.json -d "{docs/examples/viewconfs,test/{view-configs,view-configs-more}}/*.json"
 
   Test:
+    needs: [Lint, Schema]
```
manzt, line 38: These actions take almost no time, and it will prevent the long test suite from running if either fail.

app/scripts/Autocomplete.jsx
```diff
@@ -371,6 +371,7 @@ class Autocomplete extends React.Component {
             this.handleKeyUp.bind(this),
             inputProps.onKeyUp && inputProps.onKeyUp.bind(this),
           )}
+          // eslint-disable-next-line jsx-a11y/role-has-required-aria-props
```
manzt, line 374: only place where this rule was ignored

app/scripts/Button.jsx
```diff
@@ -7,7 +7,7 @@ const Button = React.forwardRef((props, ref) => (
   <button
     ref={ref}
     className={classes[props.styleName] ?? classes.button}
-    disable={props.disable}
+    disabled={props.disable}
```
manzt, line 10: not a valid prop previously

.eslintrc.js
```diff
@@ -0,0 +1,72 @@
+module.exports = {
```
pkerpedjiev, line 1: Why is this in a JS file now?

pkerpedjiev (approved):
Yay!

.eslintrc.js
manzt: Just what `npm init @eslint/config` defaults to, I think.

manzt:
I'm going to merge and open a separate PR to run prettier/eslint.

PR #1138, opened 2023-01-20
fix: compile away env vars in un-minified hglib.js

manzt:
## Description

Makes sure esbuild compiles away `process.env.NODE_ENV` environment variables in `hglib.js`. This happened previously in `hglib.min.js` since `minify: true` sets this behavior, but we need to do the same for the unminified build.

> Why is it necessary?

loading the unminifed `hglib.js` currently throws in the browser because `process` is not defined.

Fixes #___

## Checklist

- [ ] Updated CHANGELOG.md

### DIFF
--- a/scripts/compile.mjs
+++ b/scripts/compile.mjs
@@ -55,7 +55,6 @@ await fs.promises.writeFile(
 
 // delete the vite now that we have the final UMDs
 await fs.promises.unlink(viteUmdOutput);
-
 // =========================================
 //       UMD demo (dist/index.html)
 // =========================================


manzt:
cc: @alexander-veit

alexander-veit:
Just tried it and it works.

alexander-veit (approved):

scripts/compile.mjs
```diff
@@ -9,30 +9,39 @@ import * as url from 'node:url';
...
+  path.resolve(__dirname, '../dist/hglib.css'),
+);
 
 // =========================================
-//        minify (dist/hglib.min.js)
+//  prod (un)minified UMD (hglib/hglib.min)
 // =========================================
 
+// we transform the contents of the vite build
+const viteUmdOutput = path.resolve(__dirname, '../dist/higlass.umd.js');
```
pkerpedjiev, line 25: Will this mess up people that use unpkg to import higlass and rely on `hglib.js`?

alexander-veit: I think `hglib.js` and `hglib.min.js` will still be available.

manzt: @alexander-veit is correct. The umd output from Vite needs extra compilation to to make `hglib.js` and `hglib.min.js` so that we are backwards compatible. These will still expose the same API for use from `unpkg`. The ESM output from Vite is what bundlers get.

manzt:
These fixes ended up in #1141

PR #1141, opened 2023-01-29
Transpile ES6 classes for plugin tracks

manzt:
This PR extends from #1138 to further modify the final (un)minified UMD builds (`hglib.js` and `hglib.min.js`) for better backwards compatibility. 

## Description

This was a painful rabbit hole to go down. By removing babel, HiGlass no longer transpiles native ES6 syntax to ES5. This means that we ship real ES6 classes instead of de-sugaring `class` syntax to functions using prototypical inheritance. 

This causes a problem for plugin tracks because [transpiled classes cannot extend native classes](https://stackoverflow.com/a/51860850/11008641). The issue is that plugin tracks are classes which _dynamically_ extend from HiGlass's registry. Most of the plugin tracks use webpack with `babel-loader`, so this can lead to a situation where plugin authors transpile their classes but the the upstream class (e.g., `HGC.tracks.BarTrack`) is a true ES6 class. Thus the transpiled plugin can end up is incompatible with `hglib.js`.

The issue can be fixed two ways:

1. All downstream plugins also migrate to ship ES6 (i.e., remove babel-loader from webpack), using true ES6 classes all the way down
2. Upstream we continue to transpile `hglib` classes to keep compatability

This PR implements the latter, but I'd strongly suggest consider moving to 1 (and a breaking change) in the future. It should be a simple as telling folks to remove `babel-loader` from webpack, but unfortunately another higlass plugin pattern I've observed is defining plugins with arrow functions and inspecting `new.target`:

```javascript
const PluginTrack = (HGC, ...args) => {
  if (!new.target) { /* ... */ }
  /* ... */
}
```

Arrow functions are another ES6 construct so the above function is typically transformed into `function PluginTrack(...)` by babel. Untransformed, the code above throws a `SyntaxError` because `new.target` is **not** valid syntax for arrow functions: only regular functions and within class constructors.

Adds a babel transformation to the `dist/hglib.js` and `dist/hglib.min.js` to transpile ES6 classes.

> Why is it necessary?

To keep backward compatability of higlass plugins. cc: @alexander-veit 

Fixes #___

## Checklist

- [ ] Updated CHANGELOG.md


alexander-veit (approved):
Thank you for looking into this. It seems to work. Plugin tracks are now compatible again with this version without modification.

pkerpedjiev (approved):

PR #1145, opened 2023-02-27
Set selected assembly on startup

alexander-veit:
## Description

- Correctly set the selected assembly on startup.
- Hide position text in genome position search box when assembly is invalid or not set.

> Why is it necessary?

Currently the genome position search box does not show the correct assembly on startup in the select element (i.e., the one from the viewconfig).

Fixes #___

## Checklist

- [x] Updated CHANGELOG.md

### DIFF
--- a/app/scripts/GenomePositionSearchBox.jsx
+++ b/app/scripts/GenomePositionSearchBox.jsx
@@ -795,11 +795,7 @@ class GenomePositionSearchBox extends React.Component {
             className={styles['genome-position-search-bar-button']}
             id={this.uid}
             onChange={this.handleAssemblySelectEvt.bind(this)}
-            value={
-              this.state.selectedAssembly
-                ? this.state.selectedAssembly
-                : '(none)'
-            }
+            value={this.state.selectedAssembly || undefined}
           >
             {assemblyMenuItems}
           </select>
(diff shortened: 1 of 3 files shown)


app/scripts/GenomePositionSearchBox.jsx
```diff
@@ -839,7 +839,7 @@ class GenomePositionSearchBox extends React.Component {
             </div>
           )}
           renderMenu={this.handleRenderMenu.bind(this)}
-          value={this.positionText}
+          value={this.state.selectedAssembly ? this.positionText : ''}
```
manzt, line 842: should the fallback be `"(none)"` like above?

app/scripts/GenomePositionSearchBox.jsx
```diff
@@ -795,7 +795,7 @@ class GenomePositionSearchBox extends React.Component {
             className={styles['genome-position-search-bar-button']}
             id={this.uid}
             onChange={this.handleAssemblySelectEvt.bind(this)}
-            title={
+            value={
               this.state.selectedAssembly
                 ? this.state.selectedAssembly
                 : '(none)'
```
manzt, line 801: Should this have the same logic as below?

`this.state.selectedAssembly ?? this.positionText ?? '(none)'`

manzt (approved):
Looks good to me. Just a minor comment about priority order.

EDIT: Oh, actually it looks like some unit tests are failing now.

app/scripts/GenomePositionSearchBox.jsx
alexander-veit: It depends what we want to show to the user in that case. Nothing (i.e. "") or "(none)". Maybe something more informative like "No valid assembly selected" would be better?

manzt (approved):
I like the more explicit message. Thanks!

PR #1150, opened 2023-04-03
[draft] refactor: refactor react dependencies

sehilyi:
## Description

Currently, `HiGlassComponent` and its several child components use React. The use of React is to support several UI components in HiGlass when an `editable` option is set to `true` in a view config. Gosling does not use such UI components, so it is not ideal to have the React dependencies in Gosling by using HiGlass.

Toward this, this PR tries to enable using the HiGlass Component without including React components on it when demanded.

Screenshot 2023-04-03 at 15 34 32

### Thoughts

- I think we can safely not use all children components of `TrackRenderer.jsx`, such as `TrackControl.jsx` and `ListWrapper.jsx`, which looks like control buttons overlaid on top of a track

```diff
<TrackRenderer ...>
-  {topTracks}
-  {leftTracks}
-  {rightTracks}
-  {bottomTracks}
-  {galleryTracks}
-  {centerTrack}
</TrackRenderer>
```

- Implementing companion functions of `TrackRenderer`, `TiledPlot`, and `HiGlassComponent` that do not use React (and use pure javascript) might be the easiest way, e.g.,

```
higlass-embed/
   - index.js
   - TrackRendererPure.js
   - TiledPlotPure.js
   - HiGlassComponentPure.js
```

- This [article](https://medium.com/@tyler.greason/hacking-reacts-jsx-into-pure-javascript-document-fragments-57bc2d20506f) seems to be helpful

## Checklist

- [ ] Set proper GitHub labels (e.g. v1.6+, ignore if you don't know)
- [ ] Unit tests added or updated
- [ ] Documentation added or updated
- [ ] Example(s) added or updated
- [ ] Update schema.json if there are changes to the viewconf JSON structure format
- [ ] Screenshot for visual changes (e.g. new tracks or UI changes)
- [ ] Updated CHANGELOG.md

PR #1161, opened 2023-10-16
feat(core): Add initial typechecking via JSDoc comments

### LINKED ISSUE #1108: Typescript Declaration Files
Has there been any discussion regarding writing Typescript declaration files for the higlass library? Having inline documentation, typing, and IDE suggestions would be immensely helpful when writing applications using this!

manzt:
This PR adds a `tsconfig.json` and begins adding type strictness to `higlass` with [TypeScript JSDoc comments](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html).

This PR originally was targeted towards restructuring the code base into a monorepo, and I quickly realized that I want a complier helping me make code changes like this. I started with one of the largest files (`TrackRenderer.jsx`) which I am hoping to decouple from React in the future.

This PR lays the foundation to **gradually make higlass a fully typed library**, which would eventually resolve #1108 and enable more confidence when refactoring the code base. Right now, I have added types to just a handful of files, with the most major being `TrackRenderer.jsx`. The process of gradually typing is as follows:

- Pick a file
- Add `// @ts-check` to the top of the file
- Run the TypeScript type-checker (`npx tsc --noEmit`).
- Fix errors (and avoid the use of `any`)

Gradual typing will allow us the time to allocate resources to strictly type the most crucial code-paths that are associated with PRs like #1150.

TODO:

- [x] Add typechecking GH Action (i.e., `npm run typecheck`)

> Why is it necessary?

Towards a more modular higlass to serve the foundation of projects like Gosling.js and other HiGlass "apps". 

Fixes #___

## Checklist

- [ ] Updated CHANGELOG.md


app/scripts/AddTrackPositionMenu.jsx
```diff
@@ -19,14 +19,20 @@ function AddTrackPositionMenu(props) {
       >
         <tbody>
           <tr style={{ height: '30px' }}>
-            <td className={classes['add-track-position-other']} />
+            <td
```
manzt, line 22: Random eslint errors. These are spacers in the layout UI element, so they should have aria-hidden.

app/scripts/CenterTrack.jsx
```diff
@@ -1,4 +1,3 @@
-// @ts-check
```
manzt, line 1: I'd used this previously for my own sanity when refactoring, but these files currently won't pass the type checker.

app/scripts/configs/tracks-info.js
```diff
@@ -682,7 +706,7 @@ export const TRACKS_INFO = [
     aliases: ['horizontal-1d-value-interval', 'vertical-1d-value-interval'],
     datatype: ['bed-value'],
     local: false,
-    orientation: ['1d-horizontal'],
+    orientation: '1d-horizontal',
```
manzt, line 709: Think this was a bug. The track renderer assumed `orientation` was a `string` for all data types. @pkerpedjiev, could `orientation` be an `Array<string>`?

manzt:
cc: @sehilyi @etowahadams

Would either of you be open / have time for a review?

etowahadams:
Wow great to see progress on this! I find the idea of this task daunting so you inspire me. Happy to review. Are there specific things you'd want a reviewer to look out for when looking over these new types? 

I'm also curious about the choice of adding JSdoc comments vs normal typescript. Do JSdoc comments serve your goals better?

manzt:
> Are there specific things you'd want a reviewer to look out for when looking over these new types?

Yes, maybe two things to get your thoughts on explicitly:

1.) The `TrackObject` definition

Right now there is an _implicit_ interface that is assumed to exist in `TrackRenderer.jsx`. I created an explicit interface, called `TrackObject`, and gradually typed it with the help of TS based on how the interface is used in the file. 

The "real" type of `TrackObject` is:

```typescript
type TrackObject = ReturnType<TrackRenderer["createTrackObject"]>;
```

Which is the union of all the `*Track` imported at the type of the file. However, since none of the Track implementations are typed, the above isn't very precise. Right now, I've made one "lie" to TS, where I type-cast (in TS, `as TrackObject`) to this type:

```javascript
      // FIXME: Should not need to lie about the return type from createTrackObject.
      const newTrackObj = /** @type {TrackObject} */ (
        this.createTrackObject(newTrackDef.track)
      );
```

For a PR like this, I really want to avoid lying to the type-checker at any point. However, I think that going forward we can either 1.) Make sure all classes implement this new interface `TrackObject` or 2.) Independently type all the Track implementations and account for the union type in `TrackRenderer.jsx`. Either way, the path forward (that I see) is to eventually type all the Tracks. BUT, I don't want to lose the information that we are lying to the type system (so that eventually gets replaced). Do you think we could do something like? (maybe with a ref to this comment?) Or do you see a different path forward?

```javascript
      // @ts-expect-error - FIXME: Should not need to lie about the return type from createTrackObject.
      /** @type {TrackObject} */ 
      const newTrackObj = this.createTrackObject(newTrackDef.track);
```

> I'm also curious about the choice of adding JSdoc comments vs normal typescript. Do JSdoc comments serve your goals better?

Good question! Mostly just wanted to avoid adding more TS tooling to the project and renaming files. It's a little more verbose, but all the changes are mostly in comments (whereas TS would require touching function defs, etc). Also using TS would require adding more linting dependencies, to handle parsing the code for ESLint. Maybe a nice middle ground is to add a `types.ts` file that we don't check with eslint, and move all the complex type definitions there, which we then import in the JSDoc:

```javascript
/** @type {import('./types').TrackObject} */
const trackObj = { ... };
```

app/scripts/AddTrackPositionMenu.jsx
manzt: Actually, these fixes should be in a separate PR (where we bump eslint). I just reset the version.

manzt:
Just to iterate my over goals of this PR:

- Touch as little current code as possible while trying to make TS happy
- Add explicit types via comments

Once this is in place, we can then start refactoring with more confidence (i.e., touch the existing code). When reviewing, please have a closer look at _changes_ vs strictly additions (which are almost all comments). There are some places where TS yelled at me and I think there are bugs, so I've tried to remove these pieces.

app/scripts/TrackRenderer.jsx
```diff
@@ -93,29 +312,52 @@ class TrackRenderer extends React.Component {
...
+    /** @type {() => void} */
     this.zoomEndedBound = this.zoomEnded.bind(this);
 
+    /** @type {string} */
     this.uid = slugid.nice();
+
+    /** @type {string} */
     this.viewUid = this.props.uid;
 
+    /** @type {unknown} */
```
manzt, line 360: For those familiar with implementing plugin tracks, this is `HGC`. We should try to strictly type this (or make is generic) for eventual TrackRenderer.

app/scripts/TrackRenderer.jsx
```diff
@@ -191,29 +450,57 @@ class TrackRenderer extends React.Component {
...
+    /** @type {Array<unknown>} */
     this.pubSubs = [];
 
     // if there's plugin tracks, they'll define new track
     // types and we'll want to use their information when
     // we look up the orientation of a track
     if (window.higlassTracksByType) {
       // Extend `TRACKS_INFO_BY_TYPE` with the configs of plugin tracks.
-      Object.keys(window.higlassTracksByType).forEach((pluginTrackType) => {
+      for (const pluginTrackType in window.higlassTracksByType) {
```
manzt, line 467: TypeScript inference much better with `for` `in` loops.

app/scripts/TrackRenderer.jsx
```diff
@@ -237,13 +524,19 @@ class TrackRenderer extends React.Component {
   }
 
   componentDidMount() {
+    if (!this.element) {
+      throw new Error('Component did not mount, this.element is not defined.');
+    }
     this.elementPos = this.element.getBoundingClientRect();
     this.elementSelection = select(this.element);
-    this.svgTrackAreaSelection = select(this.svgTrackArea);
```
manzt, line 242: `this.svgTrackArea` does not exist in this file and there are no references across the code base. Same with `svgTrackAreaSelection`.

@pkerpedjiev

app/scripts/TrackRenderer.jsx
```diff
@@ -365,7 +663,8 @@ class TrackRenderer extends React.Component {
       const trackObject = this.trackDefObjects[track.track.uid].trackObject;
       trackObject.rerender(options);
 
-      if (track.track.hasOwnProperty('contents')) {
+      if (isCombinedTrackConfig(track.track)) {
```
manzt, line 666: This special handling of combined tracks is all over. I tried to make a type utility that we can use consistently, and will effectively narrow the type for typescript.

app/scripts/TrackRenderer.jsx
```diff
@@ -949,21 +1282,23 @@ class TrackRenderer extends React.Component {
...
 
-  updateMetaTracks() {
+  /** @param {unknown} _unused */
+  updateMetaTracks(_unused) {
     // Nothing
   }
 
+  /** @param {Array<TrackDefinition>} newTrackDefs */
   updateExistingTrackDefs(newTrackDefs) {
-    for (let i = 0; i < newTrackDefs.length; i++) {
```
manzt, line 957: Changes not totally necessary here, but keeping track of `i` in the for loop is a bit mindbending.

app/scripts/TrackRenderer.jsx
```diff
@@ -1024,14 +1362,17 @@ class TrackRenderer extends React.Component {
     return updated;
   }
 
+  /** @param {string[]} trackIds */
   removeMetaTracks(trackIds) {
     trackIds.forEach((id) => {
       this.metaTracks[id].trackObject.remove();
+      // @ts-expect-error - We are deleting the track object here
       this.metaTracks[id] = undefined;
```
manzt, line 1370: I don't this this is strictly necessary if we `delete` the key in the next line, which will also remove references to the object.

app/scripts/TrackRenderer.jsx
```diff
@@ -1088,18 +1430,21 @@ class TrackRenderer extends React.Component {
         .scale(k);
 
       this.zoomTransform = newTransform;
-      this.emptyZoomBehavior.transform(this.elementSelection, newTransform);
+      if (this.elementSelection) {
```
manzt, line 1433: could be undefined

app/scripts/TrackRenderer.jsx
```diff
@@ -1119,21 +1464,37 @@ class TrackRenderer extends React.Component {
...
+      return;
+    }
     // mouse wheel from zoom event
     // const cp = pointer(event.sourceEvent, this.props.canvasElement);
     for (const track of this.getTracksAtPosition(...this.zoomStartPos)) {
       track.movedY(movement);
     }
 
-    this.zoomTransform = this.zoomStartTransform;
+    if (this.zoomStartTransform) this.zoomTransform = this.zoomStartTransform;
```
manzt, line 1478: Same, could be undefined. Maybe we just set to `this.zoomTransform` in the constructor?

app/scripts/TrackRenderer.jsx
```diff
@@ -1627,7 +2017,7 @@ class TrackRenderer extends React.Component {
           context.setDomainsCallback = track.setDomainsCallback;
           return new ViewportTracker2D(context, options);
         }
-        return new Track(context, options);
```
manzt, line 1630: TS found. Track constructor only takes one argument.

app/scripts/TrackRenderer.jsx
```diff
@@ -1958,15 +2351,23 @@ class TrackRenderer extends React.Component {
...
 
-  forwardEvent(e) {
-    e.sourceUid = this.uid;
-    e.forwarded = true;
-    this.props.pubSub.publish('app.event', e);
+  /**
+   * Publishes an event to the pubSub channel, first overriding the
+   * sourceUid to be the uid of this track renderer.
+   *
+   * @param {any} event
```
manzt, line 2362: Should be the only explicit `any` in this PR. Maybe we could switch to `object`? But this function basically mutates any argument with additional fields and publishes.

manzt: Removed, with more accurate type.

app/scripts/TrackRenderer.jsx
```diff
@@ -85,6 +86,218 @@ const { getDataFetcher } = AVAILABLE_FOR_PLUGINS.dataFetchers;
...
+ * @property {[number, number]} position
+ * @property {[number, number]} dimensions
+ * @property {(contents: Array<TrackConfig>, x: unknown) => TrackObject} updateContents
+ * @property {ZoomedFunction} zoomed
+ * @property {(position: [number, number]) => void} setPosition
+ * @property {(position: [number, number]) => void} setDimensions
+ * @property {() => void} remove
+ * @property {(extent: number) => void} movedY
+ * @property {(yPosition: number, wheelDelta: number) => void} zoomedY
+ */
```
etowahadams, line 112: I imagine that you arrived at these properties based on what properties of `TrackObject` were being used in `TrackRenderer`? I see that some of these properties are only used in specific`*Track` classes. For example, `delayDrawing` is used in `PixiTrack` but not `SVGTrack`.

app/scripts/TrackRenderer.jsx
etowahadams: Agreed strictly typing this would be so helpful for plugin tracks! ([Link](https://github.com/gosling-lang/gosling.js/blob/f17e69f8ea9d83e9c574e864b7dfca94a49fa328/src/missing-types.d.ts#L6-L11) to the HGC type in Gosling for reference)

app/scripts/TrackRenderer.jsx
etowahadams: `CombinedTrack` is a strange track. It is the only Track doesn't extend `Track` that I can tell.

manzt: yes, my understanding is that it’s a container for multiple inner tracks.

etowahadams:
> However, I think that going forward we can either 1.) Make sure all classes implement this new interface TrackObject or 2.) Independently type all the Track implementations and account for the union type in TrackRenderer.jsx. Either way, the path forward (that I see) is to eventually type all the Tracks. BUT, I don't want to lose the information that we are lying to the type system (so that eventually gets replaced). Do you think we could do something like? (maybe with a ref to this comment?) Or do you see a different path forward?

I see the conundrum. Lie to TrackRenderer that there is a uniform interface at the risk of more difficult refactoring in the future. I agree that the long term path forward would be to eventually type all of the tracks. I'm guessing hesitation with typing the different `*Track` classes is that getting the type narrowing to behave properly in TrackRenderer would be quite burdensome. And rather than consulting the track type in `TrackConfig` to determine which `TrackObject` properties to access, things would be refactored to check the `*Track` object type directly I would imagine? 

In the short term, I guess the answer is how much the "lie" impedes the goal of decoupling from React (assuming that is the goal, along with making a monorepo). More specifically, the question seems to be how much `*Track` class specific logic is tightly coupled with React, such that refactoring without knowing `*Track` class specific properties becomes intractable. I don't have an answer to this. My impression is that the react specific functions do some of the heavy lifting and some track specific checks but I'll have to internalize the code in my head more to understand how difficult the refactor would be. 

> I'm not sure if you had any thoughts on how to handle these data types more elegantly.

I can only think of getters like you've been doing, or smart default init values. I'm guessing for many of these its not clear what a good initialization would be. I wish I could think of a more elegant way!

> Mostly just wanted to avoid adding more TS tooling to the project and renaming files. It's a little more verbose, but all the changes are mostly in comments (whereas TS would require touching function defs, etc). Also using TS would require adding more linting dependencies, to handle parsing the code for ESLint

That makes a lot of sense! I like easy opt out as a feature. I didn't realize there was such good tooling for type inference with just these JSDoc comments. Good to know.

manzt:
> And rather than consulting the track type in TrackConfig to determine which TrackObject properties to access, things would be refactored to check the *Track object type directly I would imagine?

At some point, the relationship between track config and the associated track object should be communicated to the type system. However, I think that is much further away from typing the tracks. 

> In the short term, I guess the answer is how much the "lie" impedes the goal of decoupling from React (assuming that is the goal, along with making a monorepo). 

Agreed. I think the "lie" is currently ok, and we can approach this from different angles. Currently we know that `TrackObject` satisfies the needs of `TrackRenderer`. We can either choose to use this interface to guide the typing of each `*Track`:

```javascript
/**
 * @implements {TrackObject}
 */
class Track {
  // ...
}
```

or conversely, use the new [`satisfies`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html) operator to allow the `*Track` interfaces to be more specific but ensure they "statisfy" the `TrackObject` in the context of the TrackRenderer.

> I didn't realize there was such good tooling for type inference with just these JSDoc comments. Good to know.

Yes, to my surprise as well! The [orange website](https://news.ycombinator.com/item?id=35891259) exploded over the Svelte project "getting rid of TypeScript" a little while ago, but really they just switched to TS via JSDoc. I think it's not well understood that TypeScript via JSDoc _is still TypeScript_. The type-checker gets all the same information about your code, it's just you need to provide type annotations through a different syntax (comments rather than `<>` and `:`).

PR #1163, opened 2023-10-19
chore: upgrade pixi

etowahadams:
## Description

Upgrade from `"pixi.js": "^5.0.3",` to `"pixi.js": "^6.3.0"`
Upgrade `@web/test-runner` and `@web/test-runner-puppeteer` because old versions don't support Pixi v6 WebGL elements. 

> Why is it necessary?

Gosling currently uses `"pixi.js": "^6.3.0"`. as a peer dependency. However, Higlass uses v5.  Package managers will yell at you for not matching peer dependency versions, this makes things confusing for fancy CDNs like https://esm.sh which will try to receive both v5 and v6. 

Luckily the [pixi.js v6 migration guide](https://github.com/pixijs/pixijs/wiki/v6-Migration-Guide) does not list any changes that HiGlass uses. Unluckily, the guide is not complete. 

Gosling has been using v6 for a while and things seem to behave properly. The only other testing I've done are running the unit tests locally and playing around in the development server. 

## Checklist

- [ ] Updated CHANGELOG.md


etowahadams:
Tests are passing locally now. @manzt or @sehilyi could you approve the workflow for these?

etowahadams:
Thanks Sehi! Ready for review now.

package.json
```diff
@@ -76,16 +76,16 @@
     "vkbeautify": "^0.99.3"
   },
   "peerDependencies": {
-    "pixi.js": "^5.0.3",
+    "pixi.js": "^6.3.0",
```
manzt, line 79: This is kind of surprising to me that this is the only pixi change. I expected also to see pixi change in our dev dependencies, but we already have pixi@6 listed there... Maybe we have already been using pixi@6 for development and testing for while?

manzt, line 79: This is kind of surprising to me that this is the only pixi change. I expected also to see pixi change in our dev dependencies, but we already have pixi@6 listed there... Maybe we have already been using pixi@6 for development and testing for while?

manzt: Like I don't see any pixi changes in the lockfile, except for `peerDependencies`.

manzt: If it really is the case that v5 works _and_ v6 works, then maybe we can do something similar to react/react-dom?

```suggestion
    "pixi.js": "^5.0.3 || ^6.3.0",
```

manzt (approved):
Looks great! Just a couple of comments.

package.json
etowahadams: > Maybe we have already been using pixi@6 for development and testing for while?

I think that's correct! It seem higlass was already using pixi v6 for dev, it just never got updated as a peer dep? Just changed so that pixi peer can be v5 or v6.

sehilyi: But, in this case, I believe the issue ([gosling.js/#977](https://github.com/gosling-lang/gosling.js/pull/977)) with the `@pixi/polyfill` would remain unaddressed?

manzt: Yes, I would assume that esm.sh would pick the latest pixi version from the available peer dependencies.

PR #1164, opened 2023-10-23
feat: Add most util type annotations

manzt:
## Description

> Why is it necessary?

Fixes #___

## Checklist

- [ ] Updated CHANGELOG.md

### DIFF
--- a/app/scripts/types.ts
+++ b/app/scripts/types.ts
@@ -1,0 +1,55 @@
+/* eslint-disable */
+
+export type TrackPosition = typeof import('./configs/primitives').TRACK_LOCATIONS[number];
+
+export type ChromInfo<Name extends string = string> = {
+    cumPositions: { id?: number; pos: number; chr: Name }[];
+    chrPositions: Record<Name, { pos: number }>;
+    chromLengths: Record<Name, number>;
+    totalLength: number;
+};
+
+export type UnknownTrackConfig = {
+  uid: string;
+  options: Record<string, unknown>;
+  type: string;
+  position: TrackPosition;
+  data?: Record<string, unknown>;
+  server: string;
+  tilesetUid: string;
+  coordSystem?: unknown;
+  x?: number;
+  y?: number;
+  chromInfoPath: string;
+  projectionXDomain?: [number, number];
+  projectionYDomain?: [number, number];
+  registerViewportChanged?: unknown;
+  removeViewportChanged?: unknown;
+  setDomainsCallback?: unknown;
+}
+
+export type CombinedTrackConfig = {
+  uid: string;
+  options: Record<string, unknown>;
+  contents: TrackConfig[];
+  type: 'combined';
+  position: TrackPosition;
+  data?: Record<string, unknown>;
+  server: string;
+  tilesetUid: string;
+  coordSystem?: unknown;
+  x?: number;
+  y?: number;
+  chromInfoPath: string;
+  projectionXDomain?: [number, number];
+  projectionYDomain?: [number, number];
+  registerViewportChanged?: unknown;
+  removeViewportChanged?: unknown;
+  setDomainsCallback?: unknown;
+}
+
+export type TrackConfig = UnknownTrackConfig | CombinedTrackConfig;
+
+export type TrackVisitor = {
+  (track: TrackConfig, position: null | TrackPosition): void;
+}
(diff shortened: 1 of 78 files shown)


app/scripts/utils/abs-to-chr.js
```diff
@@ -9,7 +22,9 @@ const absToChr = (absPosition, chromInfo) => {
   const lastChr = chromInfo.cumPositions[chromInfo.cumPositions.length - 1].chr;
   const lastLength = chromInfo.chromLengths[lastChr];
 
-  insertPoint -= insertPoint > 0 && 1;
+  if (insertPoint > 0) {
```
manzt, line 25: Replaced for clarity.

app/scripts/types.ts
```diff
@@ -0,0 +1,55 @@
+/* eslint-disable */
+
```
manzt, line 2: Created this file to consolidate shared types. Also nicer to do advanced TS things outside of JSDoc comments.

app/scripts/utils/clone-event.js
```diff
@@ -1,10 +1,17 @@
...
- * @param   {object}  event  Source event to be cloned.
- * @return  {object}  Cloned event
+ * Clone an event by invoking the source event's constructor and passing in the source event.
+ *
+ * @template {Event & { sourceUid?: string, forwarded?: boolean }} TEvent
+ * @param {TEvent} event - Source event to be cloned.
+ * @return {TEvent} Cloned event
  */
 const cloneEvent = (event) => {
+  if (!(event instanceof Event)) {
```
manzt, line 11: This type of cloning only works if the `event` is really an event (not just the "event" interface)

app/scripts/utils/debounce.js
```diff
@@ -9,21 +11,24 @@
  * Once webpack 2 with tree-shaking is supported I'd advocate to use lodash's
```
manzt, line 11: We are no longer using webpack, and I'm thinking we could replace this with [`just-debounce-it`](https://github.com/angus-c/just).

app/scripts/utils/dict-from-tuples.js
```diff
@@ -1,15 +1,23 @@
+// @ts-check
+
 /**
- * Create a dictionary from a list of [key,value] pairs.
- * @param tuples: A list of [key,value] pairs
- * @return: A dictionary
+ * Create a dictionary from a list of [key, value] pairs.
+ *
+ * TODO(Trevor): Replace with Object.fromEntries?
```
manzt, line 6: Think we could probably replace these implementations with aliases for:

- `dictItems`: `Object.entries`
- `dictKeys`: `Object.keys`
- `dictValues`: `Object.values`

app/scripts/utils/download.js
```diff
@@ -6,17 +13,13 @@ export function download(filename, stringOrBlob) {
     typeof stringOrBlob === 'string'
       ? new Blob([stringOrBlob], { type: 'application/octet-stream' })
       : stringOrBlob;
-  if (window.navigator.msSaveOrOpenBlob) {
```
manzt, line 9: IE stuff. Think we can cut this...

app/scripts/utils/get-element-dim.js
```diff
@@ -1,5 +1,10 @@
 import { ElementResizeListener } from '../services';
 
+/**
+ * Get the dimensions of an element
+ * @param {HTMLElement} element - DOM element to get dimensions of.
+ * @return {[width: number, height: number]} The width and height of the element.
+ */
 const getElementDim = (element) => {
   ElementResizeListener.listen();
```
manzt, line 9: I was really surprised to find this global side-effect in this "low-level" function.

app/scripts/utils/reduce.js
```diff
@@ -1,9 +1,9 @@
 /**
- * Pure functional reducer. Can be used for currying stuff. Check out
- *   `flatten.js`.
+ * Pure functional reducer. Can be used for currying stuff. Check out `flatten.js`.
  * @param   {function}  f  Reducer function.
  * @return  {array}  Curried function that accepts an array to be reduced.
  */
 const reduce = (f) => (x) => Array.prototype.reduce.call(x, f);
 
+// TODO(Trevor): Not referenced anywhere. Remove?
```
manzt, line 8: This and `flatten` are incorrectly implemented and not referenced anywhere. I'm thinking we can remove.

PR #1170, opened 2023-10-26
chore: remove unused FetchMockHelper and replace with client cache

manzt:
## Description

Extends from #1118 

> Why is it necessary?

Fixes #___

## Checklist

- [ ] Updated CHANGELOG.md


manzt:
Ok @pkerpedjiev, I now delete the `s` query param in the cache key and that seems to avoid expanding the cache.

manzt:
Ah ha! I figured out why we have tests failing interminably: Some tests rely on awaiting `requestsInFlight` but the testing framework allows stuff to run _concurrently_. That means different tests can modify "requestsInFlight" and mess each other up. 

I fixed this by setting `concurrency` to `1` in `web-test-runner.config.mjs`. Ideally we could set the requestsinflight in isolation for each of these.

manzt:
I'm going to move forward with this PR because it remove a lot of cruft. The client-side cache is experimental and can be enabled with an environment variable:

```sh
HIGLASS_USE_CACHE=true npm run test
```

PR #1171, opened 2023-10-26
feat: entrypoint for higlass/dist contents 

etowahadams:
## Description

Adds an entrypoint `hglib.css`

> Why is it necessary?

Now that higlass is behaving properly as a ES module, vite yells at you if you try to access 'higlass/dist/hglib.css'. 
```
[ERROR] Missing "./dist/hglib.css" specifier in "higlass" package [plugin vite:dep-scan]

    editor/index.tsx:6:7:
      6 │ import 'higlass/dist/hglib.css';
 ```
We could also come up with another name for this entry point. Perhaps just "./css" or something. 

## Checklist

- [ ] Updated CHANGELOG.md

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,9 +11,7 @@
     ".": {
       "import": "./dist/higlass.mjs"
     },
-    "./dist/hglib.css": {
-      "import": "./dist/hglib.css"
-    }
+    "./dist/*": "./dist/*"
   },
   "files": [
     "app",
--- a/package.json
+++ b/package.json
@@ -10,6 +10,9 @@
   "exports": {
     ".": {
       "import": "./dist/higlass.mjs"
+    },
+    "./dist/hglib.css": {
+      "import": "./dist/hglib.css"
     }
   },
   "files": [

### CODE REVIEW

**Comment by etowahadams** [comment]
Just realized I probably should cut a branch from master.

**Comment by manzt** [comment]
> Just realized I probably should cut a branch from master.

We do everything off of `develop`. I've been thinking we should remove `master`.

File: package.json [hunk 1]
```diff
@@ -10,6 +10,9 @@
   "exports": {
     ".": {
       "import": "./dist/higlass.mjs"
+    },
+    "./dist/hglib.css": {
+      "import": "./dist/hglib.css"
     }
```
manzt, line 16: I forgot about this. I think we could keep compatibility with existing bundlers with:

```suggestion
    "./dist/*": "./dist/*"
```

manzt (commented):
Good catch. Think we could allow for reaching into "./dist" with a catch all `*`.

etowahadams:
> Think we could allow for reaching into "./dist" with a catch all *.

Makes sense! Just changed.

PR #1180, opened 2023-10-28
chore(typing): opt-out of typechecking rather than opt-in

manzt:
## Description

Previously we needed to opt-in to type-checking a file, rather than opting out. This PR configures TS to type-check _all_ JS files by default, and requires `// @ts-nocheck` comment to turn off TS for file.

- Configures TS to type-check all JS files by default (`checkJs: true`)
- Removes `// @ts-check` from all previously migrated files
- Adds `// @ts-nocheck` to files that need migration

> Why is it necessary?

I think this better aligns with the long term goal of type-checking the code base, since migrating files now entails _removing_ the comment from the top of the file rather than adding it. Importantly, any new JS file will be typechecked.

## Checklist

- [ ] Updated CHANGELOG.md


manzt:
FYI: @etowahadams

PR #1182, opened 2023-10-28
chore(build): Update to latest build deps

manzt:
## Description

Bumps the versions for all things required by our build:

- esbuild
- Vite
- babel

> Why is it necessary?

Fixes #___

## Checklist

- [ ] Updated CHANGELOG.md

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+legacy-peer-deps=true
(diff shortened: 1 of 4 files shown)


.npmrc
```diff
@@ -0,0 +1 @@
+legacy-peer-deps=true
```
manzt, line 1: @alexander-veit - This will automatically use `--legacy-peer-deps` for all `npm install` commands so we don't need to remember to explicitly use it.

PR #1186, opened 2023-11-28
feat: Inject CSS via JS in final HiGlass build

manzt:
Historically users need to require both `hglib.css` and `hglib.js` in their HTML templates and bundler code. 

```html
    <link rel="stylesheet" href="https://unpkg.com/higlass@1.13/dist/hglib.css">
    <script src="https://unpkg.com/higlass@1.13/dist/hglib.min.js"></script>
```

This requires that users keep the versions in sync over time in the HTML template. 

This PR injects the final CSS via JS, eliminating the need to also have `hglib.css` in the template:

```diff
--  <link rel="stylesheet" href="https://unpkg.com/higlass@1.13/dist/hglib.css">
    <script src="https://unpkg.com/react@17/umd/react.production.min.js"></script>
    <script src="https://unpkg.com/react-dom@17/umd/react-dom.production.min.js"></script>
    <script src="https://unpkg.com/pixi.js@6/dist/browser/pixi.min.js"></script>
    <script src="https://unpkg.com/higlass@1.13/dist/hglib.min.js"></script>
```

Many JS libraries put their final CSS in JS so that loading the JS also accomplishes adding the styles to the page. 

This technique makes it easier to use use HiGlass, especially via ESM:

```javascript
import * as hglib from "http://esm.sh/higlass@1.13";
```

and ensures anytime a specific HiGlass version is imported you also have the corresponding styles for that build. It is also beneficial for libraries using HiGlass (like Gosling), since it's one less import that we need to worry about including. 

## Try it out

```
npm run build
http-server dist # open http://localhost:8080 and http://localhost:8080/esm.html
```

## Alternatives

Alternatively we keep the stylesheet separate and provide the end user with a function to load the styles via JS.

```javascript
import * as hglib from "https://esm.sh/higlass@1.13"; // doesn't load styles
await hglib.loadStyles(); // fetches ./hglib.css and adds the styles to the current document
```

Not sure what you think about these approaches @etowahadams

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Release notes
 
+## In Progress
+
+- Inject HiGlass CSS via JS, removing the need to include `hglib.css` in HTML templates.
+
 ## v1.13.2
 
 - Add entrypoint for importing from `higlass/dist/*` with bundlers
--- a/package-lock.json
+++ b/package-lock.json
@@ -92,7 +92,8 @@
         "tinyspy": "^1.0.2",
         "typescript": "^5.2.2",
         "vite": "^5.0.3",
-        "vite-plugin-babel": "^1.1.3"
+        "vite-plugin-babel": "^1.1.3",
+        "vite-plugin-css-injected-by-js": "^3.3.0"
       },
       "engines": {
         "node": ">=0.12.0"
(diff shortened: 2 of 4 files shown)


etowahadams:
Nice, I like the CSS injection! I can't think of any reasons why higlass would be used without the CSS, or with a modified CSS, especially since much of the styling is doing at the canvas level in PIXI. It seems typical for javascript libraries like plotly to not have users add a separate external stylesheet. (side note: I couldn't figure out how exactly plotly styles things from my 2 minutes of poking around but this [scss to css to js](https://github.com/plotly/plotly.js/blob/336855275a09397586731472deeecaa1d77df584/tasks/preprocess.js#L17-L27) pattern is something I've never encountered before) 

I didn't notice the dreaded FOUC either which is good.

PR #1194, opened 2024-02-16
feat: Add abstract TileSource

manzt:
## Description

Extracts a `TileSource` interface to represent an abstract "source" for HiGlass's `DataFetcher`. The `DataFetcher` does a lot of processing of the values returned from the server, so it's challenging to just implement a different source that is not a webserver (e.g., a websocket). I think there is a lot more logic that could be reused to optimize requests, but this seems sufficient for the time being.

> Why is it necessary?

This let's us connect `DataFetcher` to non-HTTP (i.e., non `fetch`-based) data sources. 

Fixes #___

## Checklist

- [ ] Updated CHANGELOG.md


manzt:
For https://github.com/higlass/higlass-python/issues/144, this would let us do something like:

```js
function JupyterDataFetcher(hgc, dataConfig, options) {
  let tileSource = { getTiles: /* ... */, getTilesetInfo: /* ... */, register: /* ... */ };
  return new hgc.DataFetcher(dataConfig, options, tileSource);
}
```

manzt:
Tbh i don't really know what `registerTileset` is or does.... I've never implemented in higlass-python or server side implementations of higlass tile sources.

pkerpedjiev:
`registerTileset` was a semi-half-baked way of fetching tilesets that were not on the server. The idea was that the client would call the `register_tileset` endpoint with a url, the server would return a uuid that the client could use to fetch tiles. It would be as if the client was fetching directly from the url without implementing the http-based tile fetching code.

I think 4DN was using it. Might be possible to deprecate.

PR #1198, opened 2024-05-30
feat: all exports Gosling uses in a single file

etowahadams:
## Description

Put all of the exports that Gosling (or the rearchitected Gosling) uses into a single file. 

Currently I call it `hggos` but would definitely be open to other names! 

``` javascript
import { hggos } from 'higlass'
const { ViewportTrackerHorizontal } = hggos

class BrushTrack extends ViewportTrackerHorizontal {

}
```

I considered several other options, but did find anything that would easily work. I'm probably missing something though, so please suggest otherwise! 

### Multiple entrypoints
I looked into what it would look like to have multiple entrypoints, and found [this](https://github.com/vitejs/vite/discussions/1736#discussioncomment-4997467) as a potential solution. 

```
// vite.config.js
lib: {
  entry: [
    path.resolve(__dirname, 'app/scripts/hglib.jsx'),
    path.resolve(__dirname, 'app/scripts/gosling.js'),
  ],
  fileName: (format, entryName) => {
    const extension = format === 'es' ? 'js' : 'cjs';
    return `js/${entryName}.${extension}`;
  },
  formats: ['umd'],
},
```

But multiple entry points are not supported when the output format is `"umd"`. So we would have to ditch the UMD build to make it work which I'm not sure we're ready to do. Or have some other more custom build. 

## Publishing multiple packages from the same repo
I also looked into what it would look like to publish a separate package. As far as I can tell this is difficult to do without turning the repo into a monorepo? Which does seem like a useful direction to go, although perhaps a bigger lift. 

Maybe there is a way to do this would going full monorepo, I'm not sure. 

## Checklist

- [ ] Updated CHANGELOG.md


app/scripts/gosling-exports.js
```diff
@@ -0,0 +1,30 @@
...
+export { default as ViewportTrackerHorizontal } from './ViewportTrackerHorizontal';
+export { default as HeatmapTiledPixiTrack } from './HeatmapTiledPixiTrack';
+/**
+ * Data Fetcher
+ */
+export { DataFetcher } from './data-fetchers';
+/**
+ * Utils
+ */
+export { fake as fakePubSub } from './hocs/with-pub-sub';
```
manzt, line 19: Curious what this is used for

app/scripts/gosling-exports.js
```diff
@@ -0,0 +1,30 @@
+// These are all exports that are
```
sehilyi, line 1:
```suggestion
// These are all exports that are
```
```suggestion
// These are all exports that are used is Gosling.js
```

manzt:
Great work! I left an idea in https://github.com/etowahadams/higlass/pull/1 for how we can progress

etowahadams:
Thanks for all the help on this! Will finish resolving conflicts and merge tomorrow 👍

PR #1206, opened 2024-12-16
Use setError for track error handling

pkerpedjiev:
## Description

Make tracks use `setError` for setting error texts. This lets one track set multiple errors by using the `source` parameter.

> Why is it necessary?

So that tracks can set and display multiple errors. If this functionality wasn't there, the different overlaps would overlap and be illegible.

Fixes #___

## Checklist

- [ ] Updated CHANGELOG.md

### DIFF
--- a/app/scripts/PixiTrack.js
+++ b/app/scripts/PixiTrack.js
@@ -315,16 +315,11 @@ class PixiTrack extends Track {
     this.errorText.x = this.position[0] + this.dimensions[0] / 2;
     this.errorText.y = this.position[1] + this.dimensions[1] / 2;
 
-    let errorTextText = '';
-    Object.values(this.errorTexts).forEach((x) => {
-      // Go through all the errors and render them.
-      if (x && x.length) {
-        if (errorTextText.length) {
-          errorTextText += '\n';
-        }
-        errorTextText += x;
-      }
-    });
+    // Collect all the error texts, filter out the ones that are empty
+    // and put the non-empty ones separate lines
+    const errorTextText = Object.values(this.errorTexts)
+      .filter((x) => x && x.length)
+      .reduce((acc, x) => (acc ? `${acc}\n${x}` : x), '');
 
     this.errorText.text = errorTextText;
     this.errorText.alpha = 0.8;
--- a/app/scripts/PixiTrack.js
+++ b/app/scripts/PixiTrack.js
@@ -208,12 +208,14 @@ class PixiTrack extends Track {
     this.errorText.anchor.x = 0.5;
     this.errorText.anchor.y = 0.5;
     this.pLabel.addChild(this.errorText);
-    /** @type {string} */
-    this.errorTextText = '';
+
     /** @type {boolean} */
     this.flipText = false;
     /** @type {import('./types').TilesetInfo | undefined} */
     this.tilesetInfo = undefined;
+
+    /** @type {{ [key: string]: string }} */
+    this.errorTexts = {};
   }
 
   setLabelText() {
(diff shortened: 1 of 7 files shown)


index.html
```diff
@@ -1,8 +1,8 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <meta charset="utf-8">
-    <meta name="viewport" content="width=device-width,initial-scale=1">
+    <meta charset="utf-8" />
```
manzt, line 4: I'm guessing these changes came from some auto formatting in the editor. I can't see to make out any actual changes to this file, so maybe we want to revert these changes.

app/scripts/TiledPixiTrack.js
```diff
@@ -224,12 +224,6 @@ class TiledPixiTrack extends PixiTrack {
     });
   }
 
-  setError(error) {
-    this.errorTextText = error;
-    this.draw();
-    this.animate();
```
manzt, line 230: Noticing that this `setError` impl that we removed has the side effect of `draw()` + `animate()` while the new base impl just calls `PixiTrack.drawError()`. Are these equivalent?

app/scripts/TiledPixiTrack.js
```diff
@@ -756,18 +773,8 @@ class TiledPixiTrack extends PixiTrack {
...
-          x.tileData && x.tileData.error && `${x.tileId}: ${x.tileData.error}`,
-      )
-      .filter((x) => x);
 
-    if (errors.length) {
-      this.errorTextText = errors.join('\n');
-    } else {
-      this.errorTextText = '';
-    }
+    this.checkForErrors();
```
manzt, line 777: Is this meant to be an inherited method? If not... I'd maybe recommend underscoring it (e.g., `_checkForErrors()`) to indicate an "internal" method.

app/scripts/PixiTrack.js
```diff
@@ -294,15 +296,43 @@ class PixiTrack extends Track {
...
+
+    this.drawError();
+  }
+
   drawError() {
     this.errorText.x = this.position[0] + this.dimensions[0] / 2;
     this.errorText.y = this.position[1] + this.dimensions[1] / 2;
 
-    this.errorText.text = this.errorTextText;
+    let errorTextText = '';
```
manzt, line 318: Small nit, took me a little bit to see what was going on here (filter + reduce). 

Not essential, but maybe we could make that more explicit. E.g.,

```javascript
const errorTextText = Object.values(this.errorTexts)
  .filter(x => x && x.length)
  .reduce((acc, x) => acc ? `${acc}\n${x}` : x, '');
```

manzt (approved):
Looks great! Some super tiny nits (feel free to ignore).

From the title, I thought something like [`Error.cause`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause) might be relevant... but now I see this isn't JavaScript error handling but rather communicating tileset errors.

index.html
pkerpedjiev: Yeah, I can checkout the original file. These are visual studio auto-formatting changes.

app/scripts/TiledPixiTrack.js
pkerpedjiev: Well the previous draw used to just call `drawError` so we're basically cutting out the middleman.

pkerpedjiev: The lack of the animate call is slightly more concerning but I've been using this function a fair amount in resglass and haven't noticed any issues so I'm inclined to leave it out.

app/scripts/PixiTrack.js
pkerpedjiev: You kids and your idiomatic expressions!

manzt (approved):
LGTM!

PR #1207, opened 2025-01-21
Switch from ESLint + Prettier to Biome

manzt:
TL;DR - Linting is now easier to maintain and takes just 5 seconds in CI, down from 5 minutes (!), thanks to replacing eslint and Prettier with the faster, all-in-one [Biome](https://biomejs.dev/).

## Description

The eslint config has proven tough to maintain over time and requires several third-party "recommended" dependencies for reasonable config. Biome is more batteries-included (and faster), and also replaces Prettier (formatting) as well as introduces a nice new feature of organizing imports.

Replaces `.eslintrc.js` with `biome.json` configuration and removes all eslint-related deps for `@biomejs/biome`. 

The first commit just replaces eslint with Biome, ignoring many rules. Subsequent commits individually enable and fix recommended rules form Biome.

- [x] Replace eslint with Biome
- [x] Replace prettier with Biome
- [x] Enable Biome's import sorting
- [x] Configure official biome GitHub action (faster install times) 

> Why is it necessary?

We incurred quite a bit of technical debt from having so much configuration for linter/formatter, which was hard to keep up over time. I think these defaults are reasonable, and a single dependency that takes care of all of this (and _fast_) can help ease this maintenance burden.

## Checklist

- [ ] Updated CHANGELOG.md


.github/workflows/ci.yml
```diff
@@ -26,17 +26,11 @@ jobs:
...
     steps:
       - uses: actions/checkout@v3
-      - uses: actions/setup-node@v3
-        with:
-          node-version: '20.x'
-          cache: 'npm'
-      - run: npm ci --legacy-peer-deps
-      - run: npm run lint
+      - uses: biomejs/setup-biome@v2 # detects version from package.json
+        run: biome ci
```
manzt, line 32: This is basically instant now in CI, since it doesn't require `npm install-ing` the project to lint. This will make identifying style errors _much_ faster.

package.json
```diff
@@ -115,15 +108,7 @@
...
-    "eslint": "^8.23.0",
-    "eslint-config-airbnb": "^19.0.4",
-    "eslint-config-prettier": "^8.5.0",
-    "eslint-plugin-import": "^2.26.0",
-    "eslint-plugin-jsx-a11y": "^6.6.1",
-    "eslint-plugin-react": "^7.31.8",
-    "eslint-plugin-react-hooks": "^4.6.0",
     "pixi.js": "^6.5.2",
-    "prettier": "^2.5.1",
     "react": "^17.0.2",
```
manzt, line 112: To me, this is the biggest benefit. We get a single dependency with reasonable/strict linting rules that takes care of all (and more) than what these 8 packages were doing for us before.

manzt:
Linting and formatting are basically instant now in CI. It's wild to me how slow the `npm ci` is for type checking and testing... and then the tests themselves are super slow because they ping against the network.

PR #1218, opened 2025-02-17
test: Mock external API requests with msw

manzt:
## Description

Many tests in this repo make external API requests to higlass servers (`higlass.io`, `resgen.io`), which are **slow and flaky**, leading to many false positives in CI.

This PR adds **mocking** for these requests using [mock service worker (msw)](https://mswjs.io/), which operates at the browser's network level via a [service worker](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API).

To try it out:

```sh
VITE_USE_MOCKS=1 npm test # first run will create `test/mocks`
VITE_USE_MOCKS=1 npm test # subsequent runs use `test/mocks`
```

## Changes

- Adds `msw` as a dev dependency.
- Implements request caching in `test/mocks/`, organizing responses by `domain` -> `route` -> `id` (e.g., `test/mocks/higlass.io/api/v1/tiles/<tileid>`).
- First test run saves API responses to disk; subsequent runs use the cache if there is a match.

There are two main files:

- `scripts/vitest-browser-commands.mjs`: server-side commands we "expose" to our mock handlers for reading/writing to the cache
- `vitest.setup.js`: behavior of msw (i.e., route handlers) for capturing/handling various requests.

## Next steps

~The cache should persist for CI and shared environments, but I'm not entirely sure how we want to do that. The full cache is ~74MB. I'm happy to check these files into the repo, but it seems not ideal for git. Another possible approach is to zip the cache, store it on S3, and fetch it in CI.~

I persisted the mocks to [higlass-test-mocks](https://github.com/higlass/higlass-test-mocks). In CI, we now clone this repo and set the `HIGLASS_MOCKS_DIR` when running the tests to test against the fixtures instead of external API requests.

## Checklist

- [ ] Updated CHANGELOG.md


test/heatmap.test.js
```diff
@@ -78,7 +78,7 @@ describe('Heatmaps', () => {
     });
   });
 
-  describe('Visualization', () => {
```
manzt, line 81: Disabled this one test which is getting back "error" data from `resgen.io`:

- `https://resgen.io/api/v1/tileset_info/?d=Bu6Djvt9T_mgZPHLig3NfQ`

```
{"Bu6Djvt9T_mgZPHLig3NfQ": {"error": "Unknown filetype ", "name": "87K x 87K pairwise genome comparison", "coordSystem": ""}}
```

vitest.setup.js
```diff
@@ -0,0 +1,259 @@
...
+    ),
+    http.get(
+      'https://s3.amazonaws.com/pkerp/public/gpsb/small.chrom.sizes',
+      () =>
+        HttpResponse.text(
+          `
+foo	249250621
+bar	243199373`.trim(),
+        ),
+    ),
```
manzt, line 253: Seems like this would be a nice win" to also avoid relying on these static files.

pkerpedjiev (approved):
Looks great!

test/heatmap.test.js
pkerpedjiev: That endpoint is fixed on resgen.io now.

Issue #1162, opened 2023-10-18
Gradual Typing of HiGlass Codebase


manzt:
This issue will serve as a tracker for our gradual transition toward a typed codebase.

## Motivation

We want to introduce type safety into HiGlass code base for improved maintainability and static error-checking. Adding types will make it easier to make planned refactors (#1150) as well provide types to end users #1108 and libraries using HiGlass (e.g. Gosling). 

## Approach

We have chosen to adopt [TypeScript via JSDoc](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html). Rather than using TypeScript syntax, you annotation your code with types via _comments_:

```typescript
// index.ts (TypeScript)
const person: Person = { name: "Trevor" };
```

```javascript
// index.js (JavaScript with comments)
/** @type {Person} */
const person  = { name: "Trevor" };
```

The TypeScript compiler can type-check both of these files the same way. Since HiGlass already includes some JSDoc annotations, using TypeScript via JSDoc seems to be the most minimal path forward in gradually introducing type-checking to the code base. In the future this could change, but currently this is the outlined approach. 

### `app/scripts/configs`
- [ ] `configs/available-track-types.js`
- [ ] `configs/colormaps.js`
- [ ] `configs/datatype-to-track-type.js`
- [ ] `configs/default-tracks-for-datatype.js`
- [ ] `configs/dense-data-extrema-config.js`
- [ ] `configs/globals.js`
- [ ] `configs/index.js`
- [ ] `configs/positions-by-datatype.js`
- [ ] `configs/primitives.js`
- [ ] `configs/README.md`
- [ ] `configs/themes.js`
- [ ] `configs/tracks-info-by-type.js`
- [ ] `configs/tracks-info.js`

### `app/scripts/data-fetchers`
- [ ] `data-fetchers/DataFetcher.js`
- [ ] `data-fetchers/genbank-fetcher.js`
- [ ] `data-fetchers/index.js`
- [ ] `data-fetchers/local-tile-fetcher.js`

### `app/scripts/hocs`
- [ ] `hocs/with-modal.jsx`
- [ ] `hocs/with-pub-sub.jsx`
- [ ] `hocs/with-theme.jsx`

### `app/scripts/plugins`
- [ ] `plugins/available-for-plugins.js`
- [ ] `plugins/get-data-fetcher.js`
- [ ] `plugins/index.js`

### `app/scripts/services`
- [ ] `services/chrom-info.js`
[...]

PR #1223, opened 2025-02-18
Remove `flatten` and `reduce` utility functions

manzt:
~These functions are not referenced any where in the code base, and the
implementation of `reduce` has a very subtle bug:~

```js
import { reduce } from "./utils/reduce.js";

const sumFrom0 = reduce((a, b) => a + b, 0);
sumFrom0([1, 2, 3, 4]); // 10

const sumFrom100 = reduce((a, b) => a + b, 100);
sumFrom100([1, 2, 3, 4]); // 10
```

~The `reduce` is missing passing the accumulator argument to the bound
function call, so instead the accumulator defaults to whatever the
_first_ element is.~

I fixed this bug in #1225. However, I think it's worth having a discussion about what to do with these and other utility functions (see below) as a part of HiGlass's public API. For example, we also export [`sum`](https://github.com/higlass/higlass/blob/4bfbfc6ef7da9eae4e0b3d4ab5d3e2ff410927f2/app/scripts/utils/sum.js#L6).

## Checklist

- [ ] Updated CHANGELOG.md


test/plugin-track.test.js
```diff
@@ -105,7 +104,6 @@ describe('Plugin track tests', () => {
       'pixiTextToSvg',
       'q',
       'rangeQuery2d',
-      'reduce',
```
manzt, line 108: This is my main concern. We export every since utility function in higlass for plugins (i.e., as a part of HiGlass's public API), so this would technically be a breaking change. 

I have no idea how much each of these functions is used in practice. I'm certain that there are some which are commonly used for higlass-specific stuff (e.g., chromsizes stuff).

manzt: A more thoughtful way of doing this would be in three parts:

1.) section off utilities which are abandoned internally (e.g., move all to a `abandoned-utils-for-plugins.js` and re-export)
2.) scrutinize our "public API" to decide what makes the cut for a breaking release
3.) make a minor release with well documented removal of various utilities

I'm not suggesting these changes for no reason. It would make it easier to achieve #1162 if we could safely "ignore" needing to worry about various utilities. Second, it would tame the API surface area of the project.

pkerpedjiev: More thoughtful way seems reasonable. Not sure I get the minor release removal. Wouldn't we just drop everything unused in v2?

Another alternative is to put these utilities into a separate `higlass-utils` package.

At first glance some of these utils are for testing purposes so they could go into a `higlass-testing` package or something to that effect?

manzt:
Closing for #1245

PR #1230, opened 2025-02-19
feat: Rewrite `tile-proxy.js`

manzt:
This was mostly an exercise to understanding what is actually going on in `tile-proxy.js`.

This PR restructures request batching in `tile-proxy.js` to improve readability and fine-tuning.

Reading through the existing code, I found it very challenging to understand the control flow. The logic is quite intertwined, with tightly coupled closures and state dependencies, making it difficult to track how requests move through the system.

My current understanding is that `DataFetcher` makes a generates a unique `id` for a tile request, dispatching a request to `fetchTilesDebounced`, which batches requests in three stages:

1.) Group by time/debouncing (groups multiple requests together)
2.) Group by `request.id`
3.) Group by `request.server`
4.) Limit max group size by threshold (`MAX_REQUEST_TILES`)

This refactor separates concerns into distinct layers:

- `bufferedBatcher` – a generic function for debounced request queueing (1)
- `fetchMultiRequestTiles` – processing batched requests
    - `bundleRequests` – consolidating requests before dispatch (2, 3, 4)

I get the "don’t fix what isn't broken" perspective, but the existing structure made it difficult to reason about what was happening.

Separating concerns improves readability, maintainability, and reusability while keeping the core logic intact.

## Checklist

- [ ] Updated CHANGELOG.md

### DIFF
--- a/app/scripts/services/tile-proxy.js
+++ b/app/scripts/services/tile-proxy.js
@@ -1,16 +1,15 @@
 import { range } from 'd3-array';
 import slugid from 'slugid';
 
-import { workerGetTiles, workerSetPix } from './worker';
+import { fetchTiles, workerSetPix } from './worker';
 
 import sleep from '../utils/timeout';
 import tts from '../utils/trim-trailing-slash';
 
-/** @import { Scale, TilesetInfo, TilesRequest }  from '../types' */
+/** @import * as types  from '../types' */
 /** @import { CompletedTileData, TileResponse, SelectedRowsOptions } from './worker' */
 
 // Config
-import { TILE_FETCH_DEBOUNCE } from '../configs/primitives';
 import { isLegacyTilesetInfo } from '../utils/type-guards';
 
 /** @type {number} */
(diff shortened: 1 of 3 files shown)


app/scripts/services/tile-proxy.js
```diff
@@ -23,273 +22,162 @@ export let requestsInFlight = 0;
...
+ * @type {(request: WithResolver<types.TilesRequest, Record<string, Tile>>) => void}
+ */
+export const fetchTilesDebounced = bufferedBatcher(fetchMultiRequestTiles);
 
-      for (const tileId of tileIds) {
-        tiles[`${data[tileId].server}/${tileId}`] = data[tileId];
+/**
+ * @param {Array<WithResolver<types.TilesRequest, Record<string, Tile>>>} requests
+ */
+async function fetchMultiRequestTiles(requests) {
```
pkerpedjiev, line 141: Docstring?

pkerpedjiev (commented):
I agree that this code was hard to understand and you call out the "if it ain't broke don't fix it" paradigm. With that said, such refactors often run the risk of substituting an implementation that was clear to one person (me) with one that is clear to another (you).

What would help mitigate that effect would be a comment somewhere in the file laying out what the current expectation of the control flow is so that if someone else has to go in to change something, they have a reference to refer to.

app/scripts/services/tile-proxy.js
```diff
@@ -753,3 +641,75 @@ const api = {
...
+  return {
+    tilesetUid: id.substring(0, firstSepIndex),
+    tileId: id.substring(firstSepIndex + 1),
+  };
+}
+
+/**
+ * Creates a buffered batch processor.
+ *
+ * Collects items and processes them in batches based on size and delay.
```
pkerpedjiev, line 657: What does it mean to process something here?

manzt:
Sorry... I'm going to close for #1231. I think we can make this refactor in a more step-wise and reviewable mannor.

manzt:
The main thing I realized is that `throttleAndDebounce` is essentially an aggregator—it transforms a function that processes a batch (`(items: Array<T>) => void`) into one that handles individual items (`(item: T) => void`), automatically batching them over time. 

The current implementation is not generic and is tightly coupled to the specific function it wraps. The first part of this PR was replacing this batch processor with something more generic, and then I got carried away (changing the implementation of the aggregator and refactoring the rest). 

I made #1231 to focus on separating this "aggregation layer", and then we can have a followup PR that does the refactor and explains the control flow.

PR #1231, opened 2025-02-20
Refactor tile proxy batching to make aggregator generic

manzt:
Although `throttleAndDebounce` has a generic name, the function signature is coupled to `fetchMultiRequestTiles`. 

For example, I would expect such function to work like:

```ts
const debounced = throttleAndDebounce((item: string) => console.log(item), 100, 100);

setInterval(() => {
  debounced("hi"); // 
}, 10);
```

However, after reviewing the implementation, I realized that `throttleAndDebounce` is actually an _aggregator_. It collects multiple calls and submits them in batches to the provided function, which operates on the aggregated input rather than individual requests.

This PR decouples it by changing the type signature, explicitly transforming a function that processes a batch `(items: Array<T>) => void` into one that handles individual items `(item: T) => void` (thus a rename to `delayedBatchExecutor`). 

```typescript
const submitItem = delayedBatchExecutor(
  (items: Array<string>) => console.log(`processing:  ${items}`), { interval: 100, finalWait: 100 }
 );

setInterval(() => {
  submitItem("hi"); // now correctly batches inputs
}, 10);
```

Notice how the function returned from `delayedBatchExecutor` operates on a scalar input rather than a batch. All the logic to handle the batch is confined to the implementation (and not the responsibility of the generic helper.

> Why is it necessary?

Fixes #___

## Checklist

- [ ] Updated CHANGELOG.md

### DIFF
--- a/app/scripts/services/tile-proxy.js
+++ b/app/scripts/services/tile-proxy.js
@@ -34,76 +34,50 @@ export let authHeader = null;
  */
 
 /**
+ * Create a function that batches calls at intervals, with a final debounce.
+ *
+ * The returned function collects individual items and executes `processBatch` at the specified interval.
+ * If additional calls occur after the last batch, a final debounce ensures they are included.
+ *
+ * @template T
  * @template {Array<unknown>} Args
  *
- * @param {(ctx: RequestContext, ...args: Args) => void} func
+ * @param {(items: Array<T>, ...args: Args) => void} processBatch
  * @param {number} interval
  * @param {number} finalWait
  */
-const throttleAndDebounce = (func, interval, finalWait) => {
+const delayedBatchExecutor = (processBatch, interval, finalWait) => {
   /** @type {ReturnType<typeof setTimeout> | undefined} */
   let timeout = undefined;
-  /** @type {Array<WrappedTilesRequest>} */
-  let bundledRequest = [];
-  /** @type {Record<string, number>} */
-  let requestMapper = {};
+  /** @type {Array<T>} */
+  let items = [];
   /** @type {number} */
   let blockedCalls = 0;
 
-  /**
-   * @param {WrappedTilesRequest} request
-   * @returns {void}
-   */
-  const bundleRequests = (request) => {
-    const requestId = requestMapper[request.id];
-
-    if (requestId && bundledRequest[requestId]) {
-      bundledRequest[requestId].ids = bundledRequest[requestId].ids.concat(
-        request.ids,
-      );
-    } else {
-      requestMapper[request.id] = bundledRequest.length;
-      bundledRequest.push(request);
-    }
-  };
-
   const reset = () => {
     timeout = undefined;
-    bundledRequest = [];
-    requestMapper = {};
+    items = [];
   };
 
-  /**
-   * @param {WrappedTilesRequest} _request
-   * @param {Args} args
-   */
-  const callFunc = (_request, ...args) => {
+  /** @param {Args} args */
+  const callFunc = (...args) => {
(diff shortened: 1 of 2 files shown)


app/scripts/services/tile-proxy.js
```diff
@@ -34,76 +34,49 @@ export let authHeader = null;
...
-   */
-  const bundleRequests = (request) => {
-    const requestId = requestMapper[request.id];
-
-    if (requestId && bundledRequest[requestId]) {
-      bundledRequest[requestId].ids = bundledRequest[requestId].ids.concat(
-        request.ids,
-      );
-    } else {
-      requestMapper[request.id] = bundledRequest.length;
```
manzt, line 65: This is the main change. This function doesn't need to do this aggregation. It's a separate concern, which we now handle in the batch processor. This allows this function to be generic.

app/scripts/services/tile-proxy.js
```diff
@@ -117,25 +91,17 @@ const throttleAndDebounce = (func, interval, finalWait) => {
...
   let wait = false;
   /**
-   * @param {WrappedTilesRequest} request
+   * @param {T} item
    * @param {Args} args
    */
-  const throttled = (request, ...args) => {
-    bundleRequests(request);
+  const throttled = (item, ...args) => {
+    items.push(item);
```
manzt, line 100: Here, we just push a new item to our queue anytime the caller makes a request.

app/scripts/services/tile-proxy.js
```diff
@@ -34,76 +34,50 @@ export let authHeader = null;
...
     // this helper function to avoid code duplication. Think of this function
     // as the actual function call that is being throttled and debounced.
-    func(
-      {
-        sessionId,
-        requests: bundledRequest,
-      },
-      ...args,
-    );
+    processBatch(items, ...args);
```
manzt, line 68: we now just submit the "bundle" (items) to a processor when either debouced or throttled.

app/scripts/services/tile-proxy.js
```diff
@@ -158,11 +124,37 @@ export const setTileProxyAuthHeader = (newHeader) => {
 export const getTileProxyAuthHeader = () => authHeader;
 
 /**
```
pkerpedjiev, line 126: Some context in the docstring here would be helpful. What does this function do? Under what circumstances? Why?

app/scripts/services/tile-proxy.js
```diff
@@ -158,11 +124,37 @@ export const setTileProxyAuthHeader = (newHeader) => {
...
- * @param {RequestContext} req
+ * @param {Array<WrappedTilesRequest>} requests
+ * @param {Array<WrappedTilesRequest>} requests
+ */
+const bundleRequests = (requests) => {
```
pkerpedjiev, line 130: Can requests be bundled by anything other than id? If not I'd consider calling this bundleRequestsById to make it clearer what is happening.

app/scripts/services/tile-proxy.js
pkerpedjiev: These (this one and the ones above) notes are helpful. I would add them as comments in the code so they're persisted.

pkerpedjiev (approved):

app/scripts/services/tile-proxy.js
manzt: I added a docstring and a unit test

app/scripts/services/tile-proxy.js
manzt: Added an comment and unit test, thanks for the suggestion!

pkerpedjiev (approved):

PR #1247, opened 2025-03-03
Bump package versions of cheerio, enzyme and react-* packages

pkerpedjiev:
## Description

Bump package versions of cheerio, enzyme and react-* packages

> Why is it necessary?

- `cheerio` was required for me for https://github.com/higlass/higlass/pull/1246
- The `react-*` packages better support React v17 and don't pull in the v16 dependency
- Enzyme is just a newer version

Fixes #___

## Checklist

- [ ] Updated CHANGELOG.md


manzt:
Curious why we needed to add cheerio if it wasn't a dep before. Something with enzyme? Let me add that change to #1246 and this can focus on react- deps

pkerpedjiev:
> Curious why we needed to add cheerio if it wasn't a dep before

Not sure. If it's not there I get this error:

```
✘ [ERROR] Missing "./lib/utils" specifier in "cheerio" package [plugin vite:dep-pre-bundle]

    node_modules/enzyme/build/Utils.js:84:21:
      84 │ var _utils = require('cheerio/lib/utils');
         ╵                      ~~~~~~~~~~~~~~~~~~~

  This error came from the "onResolve" callback registered here:

    node_modules/vite/node_modules/esbuild/lib/main.js:1150:20:
      1150 │       let promise = setup({
```

test/api.test.js
```diff
@@ -165,6 +176,8 @@ describe('API Tests', () => {
         editable: false,
       });
 
+      console.log('simpleCenterViewConfig', simpleCenterViewConfig);
```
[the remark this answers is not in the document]
manzt: For example, I switched to using a snapshot test here for out "public" API: https://github.com/higlass/higlass/blob/9cfaa7d5a416d35b1b9acb2bcb3655077381c60d/test/plugin-track.test.js#L38-L118

If we accidentally removed something, the test would fail, but if intentionally then it's just updating the snapshot (prompted by vitest with `u` command), and check in the new snapshot.

test/api.test.js
```diff
@@ -63,20 +64,26 @@ describe('API Tests', () => {
...
       const tiledPlotEl = div.querySelector('.tiled-plot-div');
       const trackRendererEl = div.querySelector('.track-renderer-div');
       const topTrackEl = div.querySelector('.top-track-container');
 
+      // Wait until these elements have stopped changing their size.
+      // Check every 20ms for up to 2 seconds
+      await waitForSizeStabilization(
+        [tiledPlotEl.parentNode, trackRendererEl, topTrackEl],
+        20,
+        2000,
```
manzt: Maybe turn these into an [options object](https://www.codereadability.com/what-are-javascript-options-objects/) (e.g., `{ timeInterval: 20, maxTime: 2000 }`)? Could probably then remove the comment above.

app/scripts/test-helpers/test-helpers.jsx
```diff
@@ -241,6 +291,132 @@ export const mountHGComponent = (
...
+ * @param {HTMLElement} div
+ */
+export const waitForComponentReady = async (div) => {
+  const elementQueries = ['.track-renderer-div'];
+  await new Promise((r) => waitForElements(div, elementQueries, r));
+
+  await waitForSizeStabilization(
+    // Check for size changes every 20 ms for 2000 seconds
+    elementQueries.map((x) => div.querySelector(x), 20, 3000),
+  );
```
manzt, line 417: Think this is a bug.

```suggestion
  await waitForSizeStabilization(
    // Check for size changes every 20 ms for 2000 seconds
    elementQueries.map((x) => div.querySelector(x)),
    20,
    3000,
  );
```

manzt: I opened https://github.com/higlass/higlass/pull/1254 against this branch which types `test-helpers.jsx` and enables type-checking so there is better auto complete, static analysis of the code here.

test/api.test.js
pkerpedjiev: Javascript needs to get its act together and allow kwargs. But in the meantime, I took your suggestion and added an options object.

This function you commented on is gone but there is another, waitForScalesStabilized which follows a similar pattern when I updated.

app/scripts/test-helpers/test-helpers.jsx
pkerpedjiev: Thanks! I just pushed some more changes and am hoping the tests finally pass (they pass locally) and then we can merge that branch in?

manzt: > merge that branch in?

sounds good to me, I can try to rebase the changes...

manzt: I rebased the changes on #1254

app/scripts/HiGlassComponent.jsx
```diff
@@ -1577,6 +1577,8 @@ class HiGlassComponent extends React.Component {
     this.xScales[uid] = xScale;
     this.yScales[uid] = yScale;
 
+    // console.trace('hsc', xScale.domain(), xScale.range());
+
```
manzt, line 1581:
```suggestion
```

app/scripts/test-helpers/test-helpers.jsx
```diff
@@ -34,40 +34,77 @@ export const changeOptions = (hgc, viewUid, trackUid, options) => {
...
-    if (trackRenderer.activeTransitions > 0) return true;
+    if (trackRenderer?.activeTransitions) return true;
   }
   return false;
 };
 
 /**
- * Wait until all transitions have finished before
- * calling the callback
+ * Waits for multiple elements to appear in the DOM and executes a callback when all are found.
```
manzt, line 53:
```suggestion
 * Waits for multiple elements to appear in the DOM.
```

package.json
```diff
@@ -61,9 +62,10 @@
     "pub-sub-es": "^2.0.1",
     "react-checkbox-tree": "^1.7.3",
     "react-color": "^2.13.8",
-    "react-grid-layout": "^0.16.6",
-    "react-simple-code-editor": "^0.9.10",
-    "react-sortable-hoc": "^1.10.1",
+    "react-grid-layout": "1.3.4",
+    "react-resizable": "^3.0.5",
```
manzt, line 66: Did we have to add a new dependency? I don't see this from before.

package.json
```diff
@@ -80,6 +82,8 @@
     "@babel/plugin-transform-classes": "^7.23.3",
     "@babel/preset-react": "^7.23.3",
     "@biomejs/biome": "^1.9.4",
+    "@testing-library/dom": "^10.4.0",
+    "@testing-library/user-event": "^14.6.1",
```
manzt, line 86: Where are these needed?

pkerpedjiev: I don't know. `npm` yelled at me that they were missing so I added them. We can try removing.

package.json
pkerpedjiev: I see something from there being imported in HiGlass.scss. Whether we need to specify it as a dependency is unclear. We can try removing it once all the tests pass.

package.json
pkerpedjiev: These are the errors I get when I remove them:

```
✘ [ERROR] Could not resolve "@testing-library/dom"

    node_modules/@testing-library/user-event/dist/esm/setup/setup.js:10:7:
      10 │ import '@testing-library/dom';
         ╵        ~~~~~~~~~~~~~~~~~~~~~~

  You can mark the path "@testing-library/dom" as external to exclude it from the bundle, which will
  remove this error and leave the unresolved path in the bundle.

✘ [ERROR] Could not resolve "@testing-library/dom"

    node_modules/@testing-library/user-event/dist/esm/document/prepareDocument.js:5:7:
      5 │ import '@testing-library/dom';
        ╵        ~~~~~~~~~~~~~~~~~~~~~~

  You can mark the path "@testing-library/dom" as external to exclude it from the bundle, which will
  remove this error and leave the unresolved path in the bundle.

✘ [ERROR] Could not resolve "@testing-library/dom"

    node_modules/@testing-library/user-event/dist/esm/setup/wrapAsync.js:1:26:
      1 │ import { getConfig } from '@testing-library/dom';
      
 ```

package.json
pkerpedjiev: Was able to remove this one.

PR #1254, opened 2025-03-09
Add type annotations for test-helpers

manzt:
## Description

> Why is it necessary?

Fixes #___

## Checklist

- [ ] Updated CHANGELOG.md


app/scripts/test-helpers/index.js
```diff
@@ -1,7 +1,5 @@
-// @ts-nocheck
 export {
   changeOptions,
-  waitForElements,
```
manzt, line 4: @pkerpedjiev I think we should avoid exporting anything new out of this module that we don't use in tests. This gives us more flexibility to make changes in the future without having to go fix tests.

PR #1304, opened 2026-04-21
Fix zoomedY not firing with mouse wheel

pkerpedjiev:
## Summary

- Check `event.sourceEvent?.shiftKey` directly in `zoomed` so value-scale zoom triggers even when Shift is pressed after a gesture has started (the cached `valueScaleZooming` flag is only set at gesture start via `zoomStarted`)
- Fall back to `deltaX` when `deltaY` is zero in `valueScaleZoom`, handling the macOS behaviour where Shift+scroll with a physical mouse wheel converts vertical scroll to horizontal

## Test plan

- [ ] Hold Shift and scroll over a 1d-horizontal track with a **trackpad** — value scale should zoom as before
- [ ] Hold Shift and scroll over a 1d-horizontal track with a **mouse wheel** — value scale should now zoom
- [ ] Confirm regular pan/zoom (no Shift) is unaffected for both input devices

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Release notes
 
+## 2.3.3
+
+- Fix value-scale zoom (`zoomedY`) not firing with a mouse wheel by checking the shift key on the current source event and falling back to `deltaX` when `deltaY` is zero
+
 ## 2.3.2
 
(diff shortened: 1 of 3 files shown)


pkerpedjiev:
Thanks for the fix! I just released v0.3.2 with your changes.

